Issue #4, opened 2018-07-12
travis redirect build unknown


styfle:
I have a repo: https://github.com/styfle/packagephobia

I tried visiting https://badgen.now.sh/travis/styfle/packagephobia which redirects to https://api.travis-ci.com/styfle/packagephobia.svg?branch=master but that shows "build unknown".

My readme is currently using https://travis-ci.org/styfle/packagephobia.svg?branch=master which redirects to https://api.travis-ci.org/styfle/packagephobia.svg?branch=master and this correctly hows "build passing"

Note the difference here is `.com` vs `.org`.

amio:
hmm, travis is in the process of [migrating .org repos to .com](https://docs.travis-ci.com/user/open-source-on-travis-ci-com/), but seems not so seamlessly.

What about adding a `/travis-org/styfle/packagephobia` for repos on travis-ci.org?

amio:
![https://badgen-service-oksomisius.now.sh/travis-org/styfle/packagephobia](https://badgen-service-oksomisius.now.sh/travis-org/styfle/packagephobia)
https://badgen-service-oksomisius.now.sh/travis-org/styfle/packagephobia

styfle:
That might work but it's kind of confusing to the end user because I don't remember off the top of my head if I am using .com or .org and which one is the open source offering. I'll probably just stick with what I have until travis finishes migrating. https://travis-ci.org/styfle/packagephobia.svg?branch=master

amio:
Sure, I'm not so satisfy with this solution either, it's a compromise(better than others comming to my mind).

It looks we can hardly get over the difference between travis .com & .org. If travis could do a 302 redirection between com & org, that would be the best solution. 😅

styfle:
This appears to be fixed now 😄

PR #9, opened 2018-07-16
Update travis badge, compat with .com & .org

amio:
@styfle I have an idea to fill the gap between travis-ci.com & travis-ci.org (#4), request both and choose the right one:

https://github.com/amio/badgen-service/blob/47ae7df3fefefd4c8e03fa7b74609a837b433714/libs/live-fns/travis.js#L4-L17


amio:
Hi @styfle I'm gonna merge it for now, if you have any thoughts on this please leave it here :D

styfle:
It looks like it works now, thanks!

PR #18, opened 2018-07-19
fix: add eslintConfig field, so IDEs are happy

olstenlarck:
A lot better.

For devs like me that only have ESLint in their editor (VSCode), it's just enough.
Also, worth nothing, because both NPM and Yarn already support "flat deps".


amio:
Yep, thanks!

PR #19, opened 2018-07-19
Update index

olstenlarck:
Adds the new things from last PRs.

We should automate this somehow... hm.

PR #20, opened 2018-07-19
fix(travis): more ensurance

olstenlarck:
Why? Because while doing #19 i've just realized that for some reason failing builds are shown as `unknown`. That so, becuase it response contains `failing` not `failed`.

I'm not sure if it depends on the different API endpoints, so.. for more safety lelts just make it regex.


amio:
I hadn't notice this, thanks!

PR #23, opened 2018-07-20
fix: updates for downloads per day

### LINKED ISSUE #22: Problem with the npm downloads/day
It's always, or mostly always, a zero.

Probably we should use the `/range` for the previous day, not for the current one. 

I can dig what Shields does, but probably is that - for the previous day.

That's problem with the API actually:
https://api.npmjs.org/downloads/point/last-day/react

In most of the time it will show 0, i assume.

olstenlarck:
resolves #22

### DIFF
--- a/libs/live-fns/npm.js
+++ b/libs/live-fns/npm.js
@@ -56,9 +56,14 @@ async function download (period, args) {
   const isTotal = period === 'total'
 
   if (isTotal) {
-    const today = new Date()
+    const now = new Date()
+    endpoint.push(`/range/2005-01-01:${now.getFullYear() + 1}-01-01`)
+  } else if (period === 'last-day') {
+    const beforeTwoDays = Date.now() - 172800000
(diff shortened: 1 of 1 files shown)


libs/live-fns/npm.js
```diff
@@ -56,9 +56,14 @@ async function download (period, args) {
...
-    const today = new Date()
+    const now = new Date()
+    endpoint.push(`/range/2005-01-01:${now.getFullYear() + 1}-01-01`)
+  } else if (period === 'last-day') {
+    const beforeTwoDays = Date.now() - 172800000
+    const dateBefore = new Date(beforeTwoDays)
+    const [date] = dateBefore.toISOString().split('T')
 
-    endpoint.push(`/range/2005-01-01:${today.getFullYear() + 1}-01-01`)
+    endpoint.push(`/range/${date}`)
```
olstenlarck: instead of this line, change it to `period = date` :D

PR #28, opened 2018-07-21
Update readme things

olstenlarck:
- Updates a bit of the wording and some typos.
- Extends the Developing section it more


package.json
```diff
@@ -6,6 +6,7 @@
   "license": "ISC",
   "scripts": {
     "lint": "standard",
+    "dev": "nodemon node service.js",
```
amio: Why is it `nodemon node service.js` other than `nodemon service.js`?

olstenlarck:
Done :D

amio:
Thanks @olstenlarck!

package.json
olstenlarck: oh.. yea, i forgot about that... :D

PR #32, opened 2018-07-24
Fix classic vs flat choices

styfle:
This changes the icons at the top to respect the style that the text says.

### Before

image

-------

### After

image

### DIFF
--- a/libs/index.md
+++ b/libs/index.md
@@ -1,8 +1,8 @@
 # Badgen
 
 Fast badge generating service.  
-[![classic](/badge/style/classic/pink)](https://badgen.net)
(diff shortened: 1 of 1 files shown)


olstenlarck:
Exactly. But what about the `flat.badgen.net`? ;d There should be the opposite in my view. :D

amio:
> respect the style that the text says.

Strongly agreed with this concern, I was think about this too, but feels putting two same style side by side looks neater. Once people click that link they will get the idea. (maybe 🤔)

Anyway this was my thoughts, what's your opinion on this explanation?

> There should be the opposite

@olstenlarck What do you mean by "opposite"?

amio:
Yep, correctness over aesthetics, merge it :D

PR #35, opened 2018-07-25
fix(npm): tweaks for dist tags and scoped packages

### LINKED ISSUE #34: Bug in badges with scoped packages
Because the "workaround" https://github.com/amio/badgen-service/blob/master/libs/live-fns/npm.js#L96

It seems that `https://registry.npmjs.org/@babel%2Fcore/*` shows empty object
but removing the star shows all versions. 

proof: https://registry.npmjs.org/@babel%2Fcore

It's interesting why with `@nestjs/core/*` it still works :laughing:

olstenlarck:
resolves #34 

Doh! :rocket: Finally works. But more and more i'm thinking to separate them as:

- `/npm/version/ava`
- `/npm/tag/ava/next` which tag will be `latest` by default, so same as above

And in reallity to show `npm version | 0.25.0` and `npm tag | 1.0.0-beta.5`, what you think?

Currently all below works:

- `/npm/v/ava` - v0.25.0
- `/npm/v/ava/latest` (same as above)
- `/npm/v/babel-core` - v6.23.x
- `/npm/v/ava/next` - v1.0.0-beta.6
- `/npm/v/next/canary` - v6.1.1-canary.2
- `/npm/v/@babel/core` - v7.0.0-beta
- `/npm/v/@babel/core/latest` (same as above)
- `/npm/v/@nestjs/core` - v5.1.0
- `/npm/v/@nestjs/core/next` - v4.6.1-1 (seems outdated)


amio:
Awesome! I'll check on this later :P

libs/live-fns/npm.js
```diff
@@ -86,16 +86,34 @@ async function download (period, args) {
...
-    : `https://registry.npmjs.org/${args}/latest`
-  const { version } = await axios.get(endpoint).then(res => res.data)
+async function npmVersion (args) {
+  const isScoped = args[0].charAt(0) === '@'
+  const name = isScoped ? [args[0], args[1]].join('%2F') : args[0]
+  let parts = null
+  let tag = args[1]
+
+  if (isScoped) {
+    parts = args[2] && args[2] !== 'latest' ? [name, args[2]] : [name]
```
amio: Might need some explanation for these stuff, endpoint examples or something, in case we forget how this comes from in the future 😄

olstenlarck: Proofs:

scoped
- PASS: https://registry.npmjs.org/@nestjs/core
- FAIL: https://registry.npmjs.org/@nestjs/core/latest
- PASS: https://registry.npmjs.org/@nestjs/core/next
So, when default tag (latest) we should request the full meta, then get `dist-tags.latest`.
Otherwise, get response data `.version`

not scoped
- PASS: https://registry.npmjs.org/ava
- PASS: https://registry.npmjs.org/ava/latest
- FAIL: https://registry.npmjs.org/ava/next

So, when default tag (latest) we should request the `/latest` enpoint, and get `.version`
Otherwise, get request the full meta search in `dist-tags` the given tag e.g `dist-tags.next`

olstenlarck: Absolute dark magic, right? :rocket:

olstenlarck:
Ah yea, forgot about that we use `npm` ;d

libs/live-fns/npm.js
amio: Absolutely 😵

How about give those parts meaningful names:
```javascript
let scope, package, tag
if (isScoped) {
	[scope, package, tag] = args
} else {
	[package, tag] = args
}
```

olstenlarck: The thing is that there is no "scope" it is joined to the `name` already.. ;d

olstenlarck: Also, we don't need and don't use them later though.

olstenlarck:
Btw, do not merge.

Should add the tag name to the label side? Like `npm@next | v5.5.2` for the `next` dist tag?

amio:
Some options:
- `npm@next | v5.5.2`
- `npm | v5.5.2@next`
- `npm | v5.5.2`(color=cyan)

🤔

olstenlarck:
Second one :+1: Cyan sounds good too, but this cyan is not cyan that i like ;d Change it to 26C6DA, 18FFFF or https://www.materialui.co/colors/cyan? There are so many better cyan colors than current one.

https://en.m.wikipedia.org/wiki/Cyan

Badge with first color:

amio:
Yeap, I pick current cyan from popular using "chat | on gitter" badge 😄

Some cyans:







My eyes... 😵

amio:
BTW, `canary` should have same color with `next`. (Found `next`, the Next.js, are using `canary` tag)

olstenlarck:
Haha, just was going to say that first 2 are good, but the third one is kinda okey too.

> BTW, canary should have same color with next. (Found next, the Next.js, are using canary tag)

yea, it won't matter what the tag name is.

amio:
How about this:

https://cdn.jsdelivr.net/npm/@nestjs/core@latest/package.json
https://cdn.jsdelivr.net/npm/@nestjs/core@beta/package.json
https://cdn.jsdelivr.net/npm/@nestjs/core@next/package.json
https://cdn.jsdelivr.net/npm/@nestjs/core@rc/package.json
https://cdn.jsdelivr.net/npm/@nestjs/core/package.json
https://cdn.jsdelivr.net/npm/webpack/package.json

olstenlarck:
Instead of the npm registry? I would prefer `unpkg.com` in that case.

amio:
Base on my human eye testing 😄 jsdelivr is faster than unpkg.com
Overall, just get rid of npm api ⚡️

olstenlarck:
I'll update the PR we get some time. I'm beating with Parcel and TypeScript ;/

amio:
No problem, take your time ☕️

olstenlarck:
Ooh, i'll get that done tomorrow. :+1:

olstenlarck:
Ooooops! I delete my fork :laughing:

PR #40, opened 2018-07-26
Add amo (mozilla add-ons) badges

tusbar:
The mozilla add-ons API (unfortunately) returns [XML](https://services.addons.mozilla.org/en-US/firefox/api/1.5/addon/markdown-viewer-chrome).

Theres is a [JSON version](https://services.addons.mozilla.org/en-US/firefox/api/1.5/addon/markdown-viewer-chrome?format=json) that does not include all the fields, such as `ratings`, `reviews` and `daily_users`.

So I’m using xml2js to transform the XML to JSON.


amio:
Cool! Thanks @tusbar

PR #45, opened 2018-07-28
Allow specifying an empty label

tusbar:
This allows things like this:

 `/circleci/github/amio/now-go?icon=circleci&label=`

to look like this:

screen shot 2018-07-28 at 3 42 34 pm

If this is something you’re willing to support. :)


amio:
Nice!

PR #46, opened 2018-07-28
Horizontally center Dependabot logo

greysteil:
Firstly, thanks for Badgen. It's awesome, and I've just switched Dependabot over to it.

I noticed that the Dependabot logo looked a little cropped on the left had side, and on close inspection found that it was slightly off-centre. This PR fixes that.


amio:
Great, thanks!

PR #47, opened 2018-07-29
add pypi!

ewdurbin:


libs/live-fns/pypi.js
```diff
@@ -0,0 +1,21 @@
...
+  switch (method) {
+    case 'v':
+      return {
+        subject: 'pypi.org',
+        status: 'v' + info.version,
+        color: info.version[0] === '0' ? 'orange' : 'blue'
+      }
+    default:
+      return {
+        subject: 'pypi',
```
olstenlarck: switch it to `pypi.org` too? for consistency? Or just make them both only `pypi`.

amio:
Great!

libs/live-fns/pypi.js
amio: +1 for both `pypi`

Like pip already did in pypi.org
https://pypi.org/project/pip/

libs/live-fns/pypi.js
```diff
@@ -0,0 +1,21 @@
+const axios = require('../axios.js')
+
+module.exports = async function (method, ...args) {
```
amio: Here we could use `package` instead of `...args` for better readability :D

libs/live-fns/pypi.js
```diff
@@ -0,0 +1,21 @@
...
+  const endpoint = `https://pypi.org/pypi/${args[0]}/json`
+  const { info } = await axios.get(endpoint).then(res => res.data)
+
+  switch (method) {
+    case 'v':
+      return {
+        subject: 'pypi.org',
```
amio: How about drop `.org`?

I saw `pip` are using `pipy | v18.0` on pipy.org https://pypi.org/project/pip/

libs/live-fns/pypi.js
ewdurbin: switched all to `pypi`

PR #49, opened 2018-07-30
fix: add patreon badge, show static usage

olstenlarck:
Why one is better? I'm still wondering which one to use in my READMEs..


olstenlarck:
aaaaah, https://github.com/amio/badgen-service/commit/4596fab0fedd3ec49399561d310bfb60caf9f958 :laughing: 

Btw, i intentionally added `fill=white` to the svg, because it seemed that it does not whitten it..

amio:
Personally I prefer 2nd, reads like the popular "chat | on gitter"

olstenlarck:
Exactly :heart:

PR #50, opened 2018-07-30
fix: npm badges and dist tags

### LINKED ISSUE #34: Bug in badges with scoped packages

olstenlarck:
fix: resolves #34 and resolves #35

### DIFF
--- a/libs/live-fns/npm.js
+++ b/libs/live-fns/npm.js
@@ -33,15 +33,33 @@ module.exports = async function npm (method, ...args) {
 }
 
 async function pkg (topic, args) {
-  const endpoint = `https://unpkg.com/${args.join('/')}/package.json`
+  let pkg = args[0]
+  let tag = '@latest'
+  let isTag = false
+
+  if (args.length === 2) {
+    pkg = args[0]
+    tag = `@${args[1]}`
+    isTag = true
+  } else if (args.length === 3) {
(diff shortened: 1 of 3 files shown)


olstenlarck:
2018-07-30-07 17 56_1280x1024_scrot

olstenlarck:
2018-07-30-07 31 55_1280x1024_scrot

libs/live-fns/npm.js
```diff
@@ -33,25 +33,64 @@ module.exports = async function npm (method, ...args) {
...
+    }
   }
 }
 
+pkg('version', ['ava'])
+pkg('version', ['ava', 'next'])
+pkg('version', ['next', 'canary'])
+pkg('version', ['babel-core'])
+pkg('version', ['@babel', 'core'])
+pkg('version', ['@nestjs', 'core', 'beta'])
```
amio: Is these for testing?

olstenlarck: Damn. I forgot them ;d But yea, that way i test the things :D

PR #51, opened 2018-07-30
Add bitbucket codecov examples

tusbar:


tusbar:
Side question: shouldn’t the label say `coverage` like coveralls’?

amio:
Thanks, and yeah, right, they should be all `coverage` or all use platform name.

tusbar:
👍 

@amio I’m also wondering if we should use the API instead of the badge.txt for codecov.

For example https://codecov.io/github/etalab/geo.data.gouv.fr/graph/badge.txt returns `4` while https://codecov.io/api/gh/etalab/geo.data.gouv.fr returns `3.9333`. 

Using the API we could also probably have more badges.

amio:
Thanks @tusbar I didn't know this api, much better than the `txt` endpoint!

PR #53, opened 2018-07-30
fix: update semver coloring

### LINKED ISSUE #52: Bug in npm badges
Again. :D 

`@babel/core` is cyan, but shouldn't be, because it is `latest`.

The cyan should be only if non-latest dist-tag is passed, no matter it is scoped or not.

olstenlarck:
resolves #52

### DIFF
--- a/libs/index.md
+++ b/libs/index.md
@@ -84,6 +84,7 @@ Available icons:
       ['version (tag)', '/npm/v/ava/next'],
       ['version (tag)', '/npm/v/next/canary'],
       ['version (scoped)', '/npm/v/@babel/core'],
+      ['version (scoped)', '/npm/v/@babel/core/latest'],
       ['version (scoped & tag)', '/npm/v/@nestjs/core/beta'],
       ['weekly downloads', '/npm/dw/express'],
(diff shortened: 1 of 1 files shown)


amio:
@olstenlarck Any idea on this topic? 😬

libs/utils/sem-color.js
```diff
@@ -10,11 +10,10 @@
  *    '0.1.2-canary.0' => 'cyan'
  */
 
-module.exports = function vc (version) {
-  if (version.match(/\b(alpha|beta|canary)\b/)) {
+module.exports = function vc (version, tag) {
+  if (tag !== 'latest') {
```
styfle: Should you check for existence here?

```js
if (tag && tag !== 'latest')
```

Otherwise I would think this code would make all npm versions cyan besides latest.

olstenlarck: Oh yea, probably the below if should be before that one.

```js
if (version.startsWith('0.')) return 'orange';
if (tag && tag !== 'latest') return 'cyan';
return 'blue';
```

That's the whole point that i'm stating in ... ah where it was. :D

PR #66, opened 2018-08-03
refactor: remove duplicate assignment on pkg

clarkdo:
@tusbar Thank you for providing such stable badge service.
Some tiny refactoring when reading the codes.


amio:
Thanks for these reviews!

tusbar:
Thanks for the nice words, but they should probably go to @amio… 😂

clarkdo:
Opps, my mistake, :smile thanks for reminding @tusbar 

@amio Thank you again

PR #67, opened 2018-08-03
Use codecov api for coverage

tusbar:
Add 1 decimal to both codecov and coveralls.


tusbar:
I don’t know what you think about the 1 decimal, but it will explain the difference in color between these too:

tusbar:
This also fixes an issue where the default branch is not `master`.

https://codecov.io/api/bb/ignitionrobotics/ign-math/branch/master

vs

https://codecov.io/api/bb/ignitionrobotics/ign-math (default branch is `default`)

amio:
Great! Thanks @tusbar

PR #71, opened 2018-08-04
Add documentation links

bokub:
I've added links to the documentation in the examples page

image

Don't hesitate to tell me if you would have designed them differently

### DIFF
--- a/libs/index.md
+++ b/libs/index.md
@@ -198,8 +198,9 @@ Available query params:
       ['yearly income', '/opencollective/yearly/webpack'],
     ]
   }
-  window.docPages = {
-    packagephobia: 'packagephobia',
+
+  window.links = {
+    packagephobia: { doc: true },
   }
 </script>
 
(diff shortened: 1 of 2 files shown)


amio:
Thanks, exactly what I planed to do!

libs/index.md
```diff
@@ -215,11 +218,13 @@ Available query params:
...
   import { html, render } from 'https://cdn.jsdelivr.net/npm/lit-html@0.10.2/lit-html.js'
 
-  const genExamples = (badges) => html`
+  const genExamples = (badges, docs) => html`
     <h4 id="live-badge">Live Badge</h4>
     <div>${Object.entries(badges).map(([service, examples]) => html`
       <dl>
-        <dt id="${service}"><a href="#${service}">${service}</a></dt>
+        <dt id="${service}"><a href="#${service}">${service}</a>
+        ${docs[service] ? html`<a href="docs/${docs[service]}" target="_blank">Docs</a>` : ''}
```
amio: I thought about using a unicode char for this, maybe `ⓘ`?

bokub: I don't know if it's visible enough.. What do you think?

image

libs/index.md
```diff
@@ -198,6 +198,9 @@ Available query params:
       ['yearly income', '/opencollective/yearly/webpack'],
     ]
   }
+  window.docPages = {
+    packagephobia: 'packagephobia',
```
tusbar: I guess this could just be an array?

tusbar: Random thought: could the future also include some sort of parameterized popin that generates complete markdown (badge + link)? 😃

libs/index.md
amio: And I thought put it right after the "packagephobia" title, maybe a `⌂` too. That's my whole plan 😂

libs/index.md
amio: Yep, a badge builder 💪

When I pulled in `lit-html`(thought about jsx before), it's on the roadmap  😄

libs/index.md
amio: Visibility won't be a problem, it's an enhancement other than core feature, once people clicked it, he will get it. I hope it won't distract too much attention.

Anyway, the feature is all good, happy to merge it 🤗
Visual things might change regularly, they are not blocking issue.

bokub:
I used a few CSS instead of a unicode character

image

amio:
Thanks @bokub 

Made a slight adjustment on style and it's live now 🎉
https://badgen.net/#packagephobia

PR #73, opened 2018-08-04
Set colors for all circleci statuses

tusbar:


amio:
Awesome! I was thinking doing this 🎉

libs/live-fns/circleci.js
```diff
@@ -10,6 +10,36 @@ module.exports = async function (vcsType, username, project, branch) {
...
+    case 'not_running':
+      return 'grey'
+
+    case 'queued':
+    case 'scheduled':
+      return 'yellow'
+
+    case 'retried':
+    case 'running':
+      return 'blue'
```
olstenlarck: I think this should be orange or somthing like that.

tusbar: I was thinking the same thing, but orange looked too much like "failed". 

Cyan wouldn’t be too bad.

amio: Yeah, right. 😅

Prefer cyan, hint some feels about "towards stable" like versions.

olstenlarck: omg

amio: Any color proposal? Or let it be cyan.

PR #76, opened 2018-08-05
feat: name, language & github badges for packagist

faustbrian:
Adding more badges for packagist and changing some labels.


libs/live-fns/packagist.js
```diff
@@ -62,13 +63,49 @@ module.exports = async function (topic, vendor, pkg, channel = 'stable') {
...
+        status: millify(response.package.github_forks),
+        color: 'green'
+      }
+    case 'ghi':
+      return {
+        subject: 'issues',
+        status: millify(response.package.github_open_issues),
+        color: 'green'
+      }
+    case 'l':
```
amio: maybe `lang` here, `l` would conflict with `license`

PR #79, opened 2018-08-06
RubyGems Support

faustbrian:
Adds support for https://rubygems.org/ which is used to publish ruby gems (packages)


amio:
Thanks @faustbrian

PR #80, opened 2018-08-06
fix(live-fns/packagist): version detection

jackmu95:
The packagist version detection for `latest` ist currently broken when a package published a patch version for a older minor release. Than this patch version is detected as the latest release (e.g.`v1.0.0` => `v2.0.0` => `v1.0.1`).

This PR adds a `compareVersions` sort function based on [compare-versions](https://github.com/omichelsen/compare-versions) to sort the versions returned from the packagist API before detecting the latest.

Other changes:
- `nodev` filters out every version containing `dev` (e.g. `1.0.x-dev`)
- `pre` filters out versions starting with `dev` (e.g. `dev-master`)
- Some minor code improvements

### DIFF
--- a/libs/utils/compare-versions.js
+++ b/libs/utils/compare-versions.js
@@ -1,0 +1,44 @@
+const indexOrEnd = (str, q) => str.indexOf(q) === -1 ? str.length : str.indexOf(q)
+const parseNumber = v => isNaN(Number(v)) ? v : Number(v)
+
+const split = (v) => {
+  var c = v.replace(/^v/, '').replace(/\+.*$/, '')
+  var patchIndex = indexOrEnd(c, '-')
+  var arr = c.substring(0, patchIndex).split('.')
+  arr.push(c.substring(patchIndex + 1))
+  return arr
+}
+
+// See: https://github.com/omichelsen/compare-versions
+module.exports = (v1, v2) => {
+  var s1 = split(v1)
+  var s2 = split(v2)
+
+  for (var i = 0; i < Math.max(s1.length - 1, s2.length - 1); i++) {
+    var n1 = parseInt(s1[i] || 0, 10)
+    var n2 = parseInt(s2[i] || 0, 10)
+
+    if (n1 > n2) return 1
+    if (n2 > n1) return -1
+  }
+
+  var sp1 = s1[s1.length - 1]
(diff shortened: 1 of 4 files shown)


libs/utils/compare-versions.js
```diff
@@ -0,0 +1,44 @@
+const indexOrEnd = (str, q) => str.indexOf(q) === -1 ? str.length : str.indexOf(q)
```
amio: Better use `compare-versions` directly other than copy it's code here

jackmu95: The problem is that it throws an exception when a invalid semver version is detected and they are quit common on packagist. This was the only reason for not using it directly.

amio: Hmm, indeed a problem, but force comparing may cause unexpected exceptions too. What about filter versions with validation before compare?

jackmu95: `semver` event provides it's own sorting function which would be ideal.
Maybe we need to sort out invalid versions first, sort them separately and put the correctly sorted (and semver valid) versions afterwards?

jackmu95: @amio I will try to come up with a initial solution later today.

amio: Sure, no hurry ☕️

jackmu95:
@amio Now the sort priority is following:
1. Non-numeric versions (e.g. `dev-master`)
2. Numeric, non-semver versions (e.g. `1.0.x-dev`)
3. Semver versions (e.g. `1.0.0`)

libs/utils/compare-versions.js
```diff
@@ -0,0 +1,39 @@
...
+
+module.exports = (v1, v2) => {
+  const validV1 = semver.valid(v1)
+  const validV2 = semver.valid(v2)
+
+  const isV1Valid = !!validV1
+  const isV2Valid = !!validV2
+
+  if (!isV1Valid && !isV2Valid) {
+    const numberVersionRegex = /^\d/
```
amio: Seems this can be omitted, since `semver.coerce()` could turn `numberVersion` into valid semver.

jackmu95:
@amio `semver.coerce()` seems like the perfect solution :)

amio:
Nice, merge it!

PR #81, opened 2018-08-06
Add brackets extension badges 🎉

bokub:
In winter 2017, I created [brackets-extension-badges.github.io](https://brackets-extension-badges.github.io/) and a [badge-provider](https://github.com/brackets-extension-badges/badge-provider-nodejs) with only one simple goal: generating badges for [Adobe brackets](http://brackets.io/) extension creators (I am one of them)

I feel like this is a good time to stop my lonely project and use **badgen** to serve that purpose.
The two projects have a few similarities: they both compute text width with a prebuilt char-width table, and both are hosted by Now Cloud ([badges.ml](https://zeit.co/bokub/badge-provider/rkmdchyvoc/source))

But there is a catch 🙁 
Brackets extension statistics cannot be retrieved using a regular API (thanks Adobe!)
All the statistics must be extracted from a big file called the 'registry', which contains everything there is to know about every extension.

You can preview this registry by running 

```sh
curl https://brackets-registry.aboutweb.com/registryList -H "Accept: application/json" -k
```
As you can see in my code, the registry needs to be parsed on startup, then at a regular interval (2 hours seems reasonable).

The interesting data is saved in memory, and retrieved on a badge request.

Because of that, 2 new dependencies are required, so I would totally understand if you declined my pull request.

Tell me what you think ! :+1:

### DIFF
--- a/libs/index.md
+++ b/libs/index.md
@@ -164,6 +164,12 @@ Available query params:
       ['version downloads', '/rubygems/dv/rails'],
       ['name', '/rubygems/n/rails'],
       ['platform', '/rubygems/p/rails'],
+    brackets: [
+      ['daily downloads', '/brackets/dd/brackets-emmet'],
+      ['weekly downloads', '/brackets/dw/brackets-emmet'],
+      ['total downloads', '/brackets/dt/brackets-emmet'],
+      ['downloads (latest)', '/brackets/dl/brackets-emmet'],
+      ['version', '/brackets/v/brackets-emmet'],
     ],
     /* CIs */
     travis: [
--- a/libs/index.md
+++ b/libs/index.md
@@ -164,6 +164,7 @@ Available query params:
       ['version downloads', '/rubygems/dv/rails'],
       ['name', '/rubygems/n/rails'],
       ['platform', '/rubygems/p/rails'],
+    ],
     brackets: [
       ['daily downloads', '/brackets/dd/brackets-emmet'],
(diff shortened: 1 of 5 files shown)


amio:
Appreciate for your trust on Badgen! ❤️

I see you already done a great work on brackets extension registry, why not continue provide it as a (unofficial) brackets extension registry? I think it's far better than registry.brackets.io. Give every extension a home page, with desc, with links, with beautiful design, like all other modern package registry does.

Let Badgen do the badge works, expose an api and that's all. ⚡️

Hope to see badges.ml come back again. (maybe under a new name like ... Unofficial Bracket Extension Registry, UBER? 😂)

amio:
Hi @bokub what's your idea on this "UBER"? [uber.now.sh](https://uber.now.sh) is still empty now :D

bokub:
Well, I'm not a fan of using the name 'UBER', it's fun for a minute but it kinda looks like cybersquatting, plus it's not great for SEO.

Currently, my idea was to fork `badgen-service` to serve **only** brackets extension badges (hosted on `badges.ml` for retro-compatibility, and some other aliases maybe), and serve a home page for each extension.

With this idea, `api.badges.ml` will automatically expose an API that could be used by `badgen.net`

bokub:
Just to let you know, [badges.ml](https://badges.ml/) is now up and working! 
The URL of the badges are retro-compatible so the previous users should not see a difference 😎

amio:
Awesome! 👍

PR #84, opened 2018-08-06
Add apm

idleberg:
`apm` is the package manager for Atom and - as far as I kow – a fork of `npm`. Hence, this PR is largely based on `npm.js`.


amio:
Nice, thanks! @idleberg

PR #97, opened 2018-08-10
core: modernize codebase and unify code style

jackmu95:
- Always use arrow functions
- Always use bracket spacing
- Avoid arrow function parentheses when possible


olstenlarck:
feels like prettier? is actually it?

definitely dont like replacing function s with const, just for tha sake of fat arrows.

Avoiding parens when possible is never a good thing. Even Airbnb recently switched to force them.

jackmu95:
@olstenlarck Yes it's prettier but everything is still StandardJS conform.

I just want to unify everything a bit more. There are currently a lot of mixed code styles.
How we unify everything is up to @amio.

libs/live-fns/amo.js
```diff
@@ -4,15 +4,21 @@ const millify = require('millify')
...
-      explicitArray: false
-    }, (err, res) => err ? reject(err) : resolve(res))
+    xml2js.parseString(
+      xml,
+      {
+        trim: true,
+        explicitArray: false
+      },
+      (err, res) => (err ? reject(err) : resolve(res))
+    )
```
amio: The original code is more readable than modified version

amio:
+1 for unify code style, the 3 rules are good too, but I think better add them as eslint rules and use `eslint --fix` to do the formatting.

Prettier has more opinionated rules, sometime the [results](https://github.com/amio/badgen-service/pull/97/files#diff-8d09f64f342314d6fbe931531ba7c7a9) are not as good as expected.

libs/live-fns/travis.js
```diff
@@ -13,12 +13,18 @@ const statuses = [
...
-    axios.get(com).then(({ data }) => data).catch(e => e),
-    axios.get(org).then(({ data }) => data).catch(e => e)
+    axios
+      .get(com)
+      .then(({ data }) => data)
+      .catch(e => e),
+    axios
+      .get(org)
+      .then(({ data }) => data)
+      .catch(e => e)
```
amio: Definitely prefer the original one here 😄

libs/serve-api.js
```diff
@@ -8,12 +8,15 @@ const serveStats = require('./serve-stats.js')
...
+  github: '240'
 }
 
 const apiHandlers = Object.entries(liveFunctions).map(([name, fn]) => {
   return get(`/${name}/*`, async (req, res) => {
-    res.setHeader('Cache-Control', `${CACHE_CONTROL}, s-maxage=${sMaxAges[name] || '120'}`)
+    res.setHeader(
+      'Cache-Control',
+      `${CACHE_CONTROL}, s-maxage=${sMaxAges[name] || '120'}`
+    )
```
amio: Prefer craft it to this
```javascript
const cacheControl = `${CACHE_CONTROL}, s-maxage=${sMaxAges[name] || '120'`
res.setHeader('Cache-Control', cacheControl)
```

jackmu95:
@amio Do you want to come up with the eslint rules or should we depend on something like `eslint-config-airbnb`?

I can even come up with a combination of `prettier` and `eslint` if you want and we can also define a precommit hook to help users commit correctly styled code.

libs/serve-stats.js
```diff
@@ -1,19 +1,15 @@
...
 const { send } = require('micro')
 
-module.exports = async function (req, res) {
-  const [
-    githubRateLimit
-  ] = await Promise.all([
-    getGithubRateLimit()
-  ])
+module.exports = async (req, res) => {
+  const [githubRateLimit] = await Promise.all([getGithubRateLimit()])
```
amio: I thought there's more to be added here, but right now the Prettier version are better for sure 😅

jackmu95:
@amio I think this is now a solid unification. I will leave the other stuff to you :)

amio:
Looks good, thanks! @jackmu95

amio:
Sometime it did have a better idea on formatting code, that's the enjoy moment :D

amio:
BTW, why do you prefer function? To me, it has more distinguishable highlighting in editor (distinguish between var & func), a subtle advantage.

olstenlarck:
Function declaration is always better thing.

amio:
ESLint is working on the peer dependencies issue: 
https://github.com/eslint/eslint/issues/10643

PR #98, opened 2018-08-10
live-fns: fix up version prefixing

jackmu95:


libs/live-fns/apm.js
```diff
@@ -33,7 +34,7 @@ async function pkg (topic, args) {
     case 'version': {
       return {
         subject: `apm`,
-        status: `v${meta.releases.latest}` || 'unknown',
+        status: meta.releases.latest ? prefixVersion(meta.releases.latest) : 'unknown',
```
amio: A bugfix :D

libs/live-fns/apm.js
```diff
@@ -1,5 +1,6 @@
 const millify = require('millify')
 const axios = require('../axios.js')
+const prefixVersion = require('../utils/prefix-version.js')
```
amio: How about `const v = require('../utils/prefix-version.js')`? Let it more handy.

libs/utils/prefix-version.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Prefix the the version with a "v".
+ *
+ * Examples
+ *   '1.2.3' => 'v1.2.3'
+ *   'v1.2.3' => 'v1.2.3'
+ *   'dev-master' => 'dev-master'
+ */
+module.exports = (version) => {
```
amio: I see this `version ? prefixVersion(version) : 'unknown'` everywhere, so I'm thinking how about make this a `version-formatter`, returns `'unknown'` while empty input, use case could be even simpler:

```javascript
const v = require('../utils/version-formatter.js')

const status = v(version)
```

jackmu95: @amio That sounds like a good solution. We can also use is everywhere then :)

libs/utils/version-formatter.js
```diff
@@ -0,0 +1,21 @@
...
+ * Formats the given version.
+ *
+ * Examples
+ *   '1.2.3' => 'v1.2.3'
+ *   'v1.2.3' => 'v1.2.3'
+ *   'dev-master' => 'dev-master',
+ *   '' => 'unknown'
+ */
+module.exports = (version) => {
+  if (!version) {
```
amio: Should we consider the case of number `0`?
Maybe `version = String(version)` at first, make it number proof.

Or add the param type info to comment block, just in case.

amio:
Thanks! @jackmu95

PR #103, opened 2018-08-11
fix(npm/downloads): try/catch axios call to return unknown

### LINKED ISSUE #102: Internal Server error on npm downloads
I understand that my package kinda don't have downloads and/or api not refreshed yet, but would be better to show `dowloads/month | unknown` or such like that.

Otherwise badges are broken.

https://github.com/olstenlarck/flow-reporter-codeframe

olstenlarck:
resolve #102


olstenlarck:
yeap

Issue #101, opened 2018-08-10
More GitHub Badges


wopian:
- [ ] Contributors
- [ ] Release downloads (total/latest)
- [ ] Last commit ![](https://img.shields.io/github/last-commit/wopian/hibari/new.svg)

wopian:
@amio can this be re-opened 😄

jneilliii:
@jackmu95, what are the requirements to get the release downloads to work?  It seems even the example on the badgen.net site is broken as it's displaying no assets.

amio:
@jneilliii fixed in 30e19d0a07c14fa5fd49ef2345b57e9c0ef3394f, thanks for reporting ❤️

PR #107, opened 2018-08-12
live-fns(github): add collaborator and contributor badges

wopian:
Completes #101.

The GraphQL API still doesn't support contributors, so that's using the V3 endpoint. Empty repos (no commits) return a 204 with an empty body, giving a length of 0.


amio:
Nice, thanks! @wopian

amio:
@wopian I removed collaborator badge in f009538725ee8cf36029daa828e8a238eecceb95 , it requires owner's token to reveal this info (ERROR 401: Must have push access to view repository collaborators.), but we shouldn't let user provide any token in badge url.

PR #108, opened 2018-08-13
live-fns: (github) refactor & use `got`

amio:
- Replace `axios` with `got`
- Separate assets downloads count, usage `/dt/` => `/assets-dl/`
- Split `stats()` into resoStats, makeRepoQuery, repoQueryBodies

Make codeclimate happy.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,11 @@
+version: "2"
+checks:
+  method-complexity:
+    config:
+      threshold: 8
+  method-lines:
+    config:
+      threshold: 100
+  file-lines:
+    config:
+      threshold: 500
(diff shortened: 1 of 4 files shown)


olstenlarck:
why got? axios is doing great and enough job. got has mdeps

amio:
No strong opinion but this [comparison](https://github.com/sindresorhus/got/issues/548#issuecomment-412008670) looks promising.It has more deps than axios but [got@9](https://packagephobia.now.sh/result?p=got) is smaller than [axios@0.18](https://packagephobia.now.sh/result?p=axios) now.

wopian:
`got` would only be a bad choice if it was being used in a client-side bundle, as its bundle size is 6x larger than [axios@0.18](https://bundlephobia.com/result?p=axios@0.18.0) (68/21kb vs 12/4kb). As of ^9, got is [no longer browser friendly](https://github.com/sindresorhus/got/issues/549#issuecomment-411300852) so that doesn't really matter.

PR #109, opened 2018-08-13
Retrieve trackingGA from env

bokub:
Right now, badgen is using a fixed Google Analytics tracking ID.
This ID is used by forks of badgen, as well as local versions running on a dev server, which tend to alter real statistics from `badgen.net`.

This ID should be used only in production, and maybe you'd like to change it later without editing your code.

Maybe you can add your tracking ID as a secret : 

```sh
now secrets add tracking-ga "UA-4646421-14"
```

### DIFF
--- a/libs/serve-docs.js
+++ b/libs/serve-docs.js
@@ -16,7 +16,7 @@ const serveMarkedOptions = {
     p img + a { vertical-align: top; margin-left: 0.4em; font: 14px/20px monospace }
   `,
   beforeHeadEnd: `<link rel="icon" type="image/svg+xml" href="/favicon.svg">`,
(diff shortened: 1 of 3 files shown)


amio:
@bokub I update the env to "@badgen-tracking-ga", in case conflict with future env vars.

bokub:
Oh no, I wouldn't realize it would raise this error...

image

Isn't there a way to make this env variable optional?

amio:
Just going to update README 😅

Use `now amio/badgen-service -e GH_TOKEN='' -e TRACKING_GA=''`

Proposed an [issue](https://github.com/zeit/now-cli/issues/1524), hope now-cli would support optional env vars.

PR #119, opened 2018-08-26
Fix uptime link

bokub:
The uptime badge says 'past week uptime' but was showing the **last day** uptime.

Also, I removed a few unused links at the end of the README


amio:
Thanks!

PR #128, opened 2018-08-29
icons: add missing icons to index.md

samlanning:
Perhaps it would be cool if this list was automatically generated from the files in that directory?


wopian:
See https://github.com/amio/badgen-service/pull/99#issuecomment-412114268 for why they're missing.

PR #99, opened 2018-08-10
docs: add all missing icons

jackmu95:

PR #129, opened 2018-08-29
icons: automatically populate list in index.md

samlanning:
After opening #128, but before seeing #99, I was thinking that it would be a good idea to automatically populate the list of icons on the homepage, rather than having to maintain the list there, and remembering to add icons each time a new one is added.

After seeing that some are deliberately not on there, I think this still could be useful now i've added a blacklist, and will self-document the fact that some badges have deliberately been chosen not to be on there, and avoid any further confusion of users introducing PRs like #128 and #99.


amio:
Thanks! @samlanning

PR #132, opened 2018-08-31
live-fns: (lgtm) use default accept header

samlanning:
Our API has been updated now, and `application/json` `Accept` headers work, so this special behaviour is no longer needed. :smile:


amio:
Awesome!

PR #136, opened 2018-09-04
Override badge color

bokub:
I'd love to be able to override a live badge color, for example having my npm download count in pink, or my license badge in some weird color I like.

This PR adds an optional 'color' option to the badge handler (live & static), which supports color names and hex strings.

E.g:  
- Badge  `/npm/dm/express?color=pink`
- Badge  `/npm/license/lodash?color=9ea7f9`


amio:
Thanks! @bokub 
This is a must have.

PR #141, opened 2018-09-05
Add Keybase support

### LINKED ISSUE #140: Keybase badge
See: https://github.com/keybase/keybase-issues/issues/849 and https://github.com/badges/shields/issues/1299

Something like: image

lukechilds:
Resolves #140 

Also indirectly resolves https://github.com/keybase/keybase-issues/issues/849

Will show the 64-bit PGP key fingerprint for the user's primary PGP key.

screen shot 2018-09-05 at 11 49 30 am

If the user doesn't exist it will fallback to a grey unknown badge.

screen shot 2018-09-05 at 11 51 25 am


lukechilds:
Using the Now deployment:

`https://badgen-service-gkdwhmmfno.now.sh/keybase/pgp/lukechilds`



`https://badgen-service-gkdwhmmfno.now.sh/keybase/pgp/nonexistentuser`

libs/live-fns/keybase.js
```diff
@@ -0,0 +1,28 @@
...
+const getPGPFingerprint = user => {
+  const fingerprint = user.public_keys.primary.key_fingerprint
+
+  // Convert to 64-bit fingerprint
+  return fingerprint.substr(fingerprint.length - 16).toUpperCase().match(/.{1,4}/g).join(' ')
+}
+
+module.exports = async (topic, username) => {
+  const endpoint = `https://keybase.io/_/api/1.0/user/lookup.json?usernames=${username}`
+  const user = await axios.get(endpoint).then(res => res.data.them[0])
```
sindresorhus:
```js
const {data} = await axios.get(endpoint)
const user = data.them[0]
```

amio:
Awesome! Thanks @lukechilds 
Since this repo use npm (package-lock.json), I removed `yarn.lock`.

libs/live-fns/keybase.js
```diff
@@ -0,0 +1,28 @@
+const axios = require('../axios.js')
```
sindresorhus: Correct me if I'm wrong, but I think `Got` is used for new services. Re: https://github.com/amio/badgen-service/commit/c9081aa8b03c66bb8f484c047758a2dd386f71f7

amio: I'm planning do the refactor myself :D

lukechilds:
Oops, sorry missed `yarn.lock`.

libs/live-fns/keybase.js
```diff
@@ -0,0 +1,29 @@
...
+
+const getPGPFingerprint = user => {
+  const fingerprint = user.public_keys.primary.key_fingerprint
+
+  // Convert to 64-bit fingerprint
+  return fingerprint.substr(fingerprint.length - 16).toUpperCase().match(/.{1,4}/g).join(' ')
+}
+
+module.exports = async (topic, username) => {
+  const endpoint = `https://keybase.io/_/api/1.0/user/lookup.json?usernames=${username}`
```
sindresorhus: Use the Got [`query`](https://github.com/sindresorhus/got#query) option with an object. The current will probably work fine since it's just a username, but best practice and all.

sindresorhus:
Use the `fields` query param to only get the field we're interested in. See: https://keybase.io/docs/api/1.0/call/user/lookup 

> A user object is a large dictionary and could contain pretty much everything about a user that you have access to. (See the user objects page in the documentation for more info.)

> To avoid getting more data than you need, you can (should!) specify a comma-separated list of fields you desire.

libs/live-fns/keybase.js
```diff
@@ -0,0 +1,29 @@
+const got = require('../got.js')
+
+const getPGPFingerprint = user => {
```
sindresorhus: I think the function should only convert the fingerprint to 64-bit. The `user.public_keys.primary.key_fingerprint` part should just be in the main function. Alternatively, it should be split into two functions.

lukechilds: Ok, I split it up like that because we need to check if `user` is null or not first. So I was planning on implementing that check, and then just passing the user object over to the function if all is good.

However, it seems like Badgen is automatically falling back to the default if anything throws so I didn't end up implementing a check.

lukechilds: You mean like this?

https://github.com/amio/badgen-service/pull/141/commits/c27be549e8f2d5d0d3112effc91137f80f3ceca6

sindresorhus: Yes, but rename the function to something like `convertFingerprintTo64bit`.

lukechilds: Done 👌

libs/live-fns/keybase.js
```diff
@@ -0,0 +1,29 @@
+const got = require('../got.js')
+
+const convertFingerprintTo64bit = fingerprint => fingerprint.slice(fingerprint.length - 16).toUpperCase().match(/.{1,4}/g).join(' ')
```
amio: This can be `.slice(-16)`

lukechilds:
@sindresorhus Good spot re fields param, done that: https://github.com/amio/badgen-service/pull/141/commits/b5c86c916dd4df44186016966e188f70b51655ef

libs/live-fns/keybase.js
lukechilds: Done https://github.com/amio/badgen-service/pull/141/commits/3d8ca72df11b1f5a92289d9dea9d8e35922a0b32

amio:
Great! Thanks @lukechilds @sindresorhus

PR #151, opened 2018-09-10
Change david-dm to green w/zero deps

styfle:
I like to add the `/david/dev` and `/david/dep` to all projects and it looks a little odd when they are not both green.

My subjective opinion is that zero dependencies should be considered "passing" (up to date) and should be green, not blue.


amio:
Make sense 👍

PR #154, opened 2018-09-13
Add xo badges

tusbar:
I like having xo badges on my repos. :D

/cc @sindresorhus

### DIFF
--- a/libs/live-fns/xo.js
+++ b/libs/live-fns/xo.js
@@ -1,0 +1,49 @@
+const got = require('../got.js')
+
+const getIndent = space => {
+  if (space === false) return 'tab'
+  if (space === true) return '2 spaces'
+  if (space === 1) return '1 space'
+  return `${space} spaces`
+}
+
+module.exports = async (topic, ...project) => {
+  const endpoint = `https://cdn.jsdelivr.net/npm/${project.join('/')}/package.json`
+  const data = await got(endpoint).then(res => res.body)
+
+  if (!data.devDependencies || !('xo' in data.devDependencies)) {
+    return {
+      subject: 'xo',
+      status: 'not enabled',
+      color: 'grey'
+    }
+  }
+
+  const {
+    semicolon = true,
+    space = false
+  } = (data.xo || {})
+
+  switch (topic) {
+    case 'status':
+      return {
+        subject: 'xo',
+        status: 'enabled',
+        color: '5ED9C7'
+      }
+
+    case 'indent':
+      return {
+        subject: 'indent',
+        status: getIndent(space),
+        color: '5ED9C7'
+      }
+
+    case 'semi':
+      return {
+        subject: 'semicolons',
+        status: semicolon ? 'enabled' : 'disabled',
+        color: '5ED9C7'
+      }
+  }
+}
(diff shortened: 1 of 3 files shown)


libs/live-fns/xo.js
```diff
@@ -0,0 +1,50 @@
...
+  if (xoConf.space === true) {
+    xoConf.space = 2
+  }
+
+  switch (topic) {
+    case 'status':
+      return {
+        subject: 'xo',
+        status: 'enabled',
+        color: 'cyan'
```
amio: Better use XO's theme color `5ED9C7`

libs/live-fns/xo.js
```diff
@@ -0,0 +1,50 @@
...
+      subject: 'xo',
+      status: 'not enabled',
+      color: 'grey'
+    }
+  }
+
+  const xoConf = {
+    semicolon: true,
+    space: false,
+
```
amio: This line can be removed

libs/live-fns/xo.js
```diff
@@ -0,0 +1,50 @@
...
+      return {
+        subject: 'xo',
+        status: 'enabled',
+        color: 'cyan'
+      }
+
+    case 'indent':
+      return {
+        subject: 'indent',
+        status: xoConf.space
```
amio: Maybe make it a standalone function?

```javascript
const getIndent = space => {
  if (space === false) return 'tab'
  if (space === true) return '2 spaces'
  if (space === 1) return '1 space'
  return `${space} spaces`
}
```

olstenlarck:
This should be some joke.

amio:
@olstenlarck It's simple but still fine and useful, so why not? We have a lot of dynamic licence badges, even I used one on [README](https://github.com/amio/badgen-service/commit/df271d2fdd3385f94ec4fbc2cb741a08970108e6#diff-04c6e90faac2675aa89e2176d2eec7d8R8) 😄

amio:
@tusbar Thanks!

styfle:
This is a good idea. Should we make a similar one for Prettier with the iconic pink? https://prettier.io

olstenlarck:
@amio license is one thing, some (someone') style is totally another.
Then we should add for @standard and SemiStandard. And as @styfle mention - Prettier.

In any way, such badges are ridiculous idea and they are not truly "dynamic" in heart. It not make any sense to have such "dynamic" thing, because you don't change or do anything related to that in every single commit. CI badges make sense because the service can provide you with feeling what is going on.

It not make sense, it just don't. It's more like a promotion/discoverability than anything.

amio:
I understand your point, thanks for sharing, while I thinks there's more perspective for this.

As an author, I know static badge has better performance than live badge, I thought I would use that on a license badge which usually won't change ever. But actually as a user, I prefer to use live badge. This gives me a thought: user don't (want to) think about performance, that should be author's concern. From a user's perspective, every badge __should__ serve the best performance, and every convenience count. That's why I choose live badge for license badge as a user.

I would love to support various use cases, as long as it doesn't cost too much. Some are must-have, some are good-to-have, and of course some are not-to-have. This case is a good-to-have as I explained previously, it won't bother me to "complete" this list (of linter badges), but if someone send a PR, why refuse that?

papb:
Does adding this functionality decreases the performance of other live badges?

olstenlarck:
> Does adding this functionality decrease the performance of other live badges?

No. Everything is separate : )

@amio yup.

amio:
@papb I talked a lot about perf (seems causing some confusion) :P, that case (static vs live) is to show convenience is more weight than other factor (like perf etc.) from a user's perspective.

papb:
Oh, I see. In this case, I think this PR is good too :)

PR #158, opened 2018-09-17
feat(github): add statuses badge

tunnckoCore:
yeaaah, closes #157 :heart_eyes: 

There should probably also have and for each status, but.. not so needed.

This allows to totally move to "Now CI" (as i call it):
https://zeit.co/blog/every-push-now

PR #161, opened 2018-09-18
Continue  #158

amio:
A clone of #158 but reverted the deletion of package-lock.json.


libs/index.md
```diff
@@ -70,6 +70,10 @@ Advanced usage (for badge makers):
       ['latest stable release', '/github/release/babel/babel/stable'],
       ['latest tag', '/github/tag/micromatch/micromatch'],
       ['watchers', '/github/watchers/micromatch/micromatch'],
+      ['status', '/github/status/micromatch/micromatch'],
+      ['status', '/github/status/tunnckoCore/badgen-service/github-status'],
```
amio: This is a "status branch" example, duplicated with "micromatch/micromatch/dev", and I guess you gonna delete it after PR merged? @tunnckoCore

libs/index.md
```diff
@@ -70,6 +70,10 @@ Advanced usage (for badge makers):
...
+      ['status branch', '/github/status/micromatch/micromatch/dev'],
```
amio: micromatch doesn't have a `dev` branch

libs/index.md
tunnckoCore: Was going to mention after i PR-ed, that i wont deleting the repo and branch.

libs/index.md
tunnckoCore: It have.. :D

tunnckoCore:
Was going to mention after i PR-ed, that i wont deleting the repo and branch.
At least, until we found some pending one? It's cool to show it. It won't be problem for me to not deleting it.

libs/index.md
amio: Isn't the dev branch this https://github.com/micromatch/micromatch/tree/dev ?

libs/index.md
amio: Not sure what the status would be if I merged this one 😂

amio:
@tunnckoCore Seems this one ([micromatch/micromatch/gh-pages](https://badgen-service-ovkqzadxac.now.sh/github/status/micromatch/micromatch/gh-pages)) is always pending

tunnckoCore:
Oh, yea probably is not found ;d;d

> is always pending.

haha, strange, but it's not guaranteed, IMO. Hm.. how we can make pending status? That's what we should learn :D THen we can add such branch here ;p

amio:
The states looks a bit weird, sometimes it shows check status (success, error), sometime it looks like merging status (pending).

But anyway, let's merge this.

PR #203, opened 2018-11-23
Add Azure Pipelines (close #200)

### LINKED ISSUE #200: Azure Pipelines support
Will this service supports Azure Pipelines? More and more GitHub repositories use Azure Pipelines now, such as Yarn and TypeScript.

g-plane:
This is my first PR to this repo, so I'm not familiar too much...

### DIFF
--- a/libs/live-fns/azure-pipelines.js
+++ b/libs/live-fns/azure-pipelines.js
@@ -9,7 +9,7 @@ module.exports = async (organization, project, definition, branch = 'master') =>
     return {
       subject: 'Azure Pipelines',
       status: 'unknown',
-      color: 'blue'
+      color: 'grey'
     }
   }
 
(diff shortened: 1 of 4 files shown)


libs/live-fns/azure-pipelines.js
```diff
@@ -0,0 +1,29 @@
...
+
+module.exports = async (organization, project, definition, branch = 'master') => {
+  const response = await got(`https://dev.azure.com/${organization}/${project}/_apis/build/status/${definition}?branchName=${branch}`, { json: false })
+  const contentType = response.headers['content-type']
+
+  if (!contentType.includes('image/svg+xml')) {
+    return {
+      subject: 'Azure Pipelines',
+      status: 'unknown',
+      color: 'blue'
```
amio: `'grey'` would be better for unknown state

g-plane:
Maybe Azure DevOps doesn't provide JSON API at all. If you check the source code in Shields.io, it also fetches the SVG and parses it.

libs/live-fns/azure-pipelines.js
g-plane: Now it was fixed.

amio:
@g-plane Hmm, just curious :D

PR #216, opened 2018-12-28
github tag got the first instead of last

kalsky:
When using the 'tag' option to get the latest tag - it gets the oldest (first) tag instead of the latest one.
This PR fixes this.


amio:
@kalsky Sorry for missing this!

PR #255, opened 2019-02-28
Update amo

tusbar:
Use new API (that returns json <3) – so get rid of xml2js.

Current badges are broken.


amio:
Great! Thanks @tusbar

Issue #268, opened 2019-03-21
Remove running and not run CircleCi status'


AndrewLeedham:
I feel that the `running` and `not run` values for CircleCI badges feel a bit redundant:

### `running`
Badges are often cached for some time, the running state often doesn't reflect the actual state as by the time you see the badge it probably isn't running anymore.

### `not run`
One of the common reasons for skipping CI is for updating version/changelog information after a release. Seeing `not run` would therefore not tell you much about what is going on in CI.

### Proposal
Personally I think it would make sense for the state to work the same as shields.io. Where it either shows `passed` or `failed`, depending on the most recent job that wasn't skipped.

Perhaps there are other use cases I am missing, where `not run` and `running` would be useful, so perhaps a separate badge for this use case would be more appropriate?

tunnckoCore:
Agree.

AndrewLeedham:
@tunnckoCore Do you think the current implementation should be modified, or introduce a new badge?

tunnckoCore:
Modified. To me it makes sense.

amio:
Sorry for the late response, this is great!

PR #269, opened 2019-03-22
live-fns: simplify circleci badge - success/failed

AndrewLeedham:
fixed: #268


amio:
@tunnckoCore Couldn't find any info in the api docs, but I guess there won't be `timedout` if the task is completed.

This looks fine, thanks! @AndrewLeedham

PR #280, opened 2019-05-27
Replaced Jekyll repository for CodeClimate's own repo

Drakota:
The CodeClimate badges for Jekyll stopped working and I think using CodeClimate's own repo would assure that this badge will always work even in the advent where a repo (like Jekyll) could stop using CodeClimate.


amio:
That's great! Thanks @Drakota

Issue #283, opened 2019-06-01
Support FOSSA and Dependabot


tunnckoCore:
# @Dependabot has 2 badges. 

One is for status, which I believe is working and dynamic (not like the Renovate's one yet)

```
https://api.dependabot.com/badges/status?host=github&repo=tunnckoCore/asia
```
![](https://api.dependabot.com/badges/status?host=github&repo=tunnckoCore/asia)

```
https://api.dependabot.com/badges/status?host=github&repo=tunnckoCore/babel-preset-esmc
```
![](https://api.dependabot.com/badges/status?host=github&repo=tunnckoCore/babel-preset-esmc)

Another one is the so-called `compatibility` score badge, but... I'm guessing it's too dynamic.

```
[![Dependabot compatibility score](https://api.dependabot.com/badges/compatibility_score?dependency-name=lodash&package-manager=npm_and_yarn&previous-version=4.17.10&new-version=4.17.11)](https://dependabot.com/compatibility-score.html?dependency-name=lodash&package-manager=npm_and_yarn&previous-version=4.17.10&new-version=4.17.11)
```

results in [![Dependabot compatibility score](https://api.dependabot.com/badges/compatibility_score?dependency-name=lodash&package-manager=npm_and_yarn&previous-version=4.17.10&new-version=4.17.11)](https://dependabot.com/compatibility-score.html?dependency-name=lodash&package-manager=npm_and_yarn&previous-version=4.17.10&new-version=4.17.11) see the page to which it leads [here](https://dependabot.com/compatibility-score/?dependency-name=lodash&package-manager=npm_and_yarn&previous-version=4.17.10&new-version=4.17.11).

_(I definitely don't like Dependabot and don't want to support and embrace its usage, but here... why not... ;d)_

---

# As about the FOSSA

It has a passing/failing badge like this, and supports github bitbucket gitlab?

```
[![FOSSA Status](https://app.fossa.io/api/projects/git%2Bgithub.com%2FtunnckoCoreLabs%2Fparse-function.svg?type=shield)](https://app.fossa.io/projects/git%2Bgithub.com%2FtunnckoCoreLabs%2Fparse-function?ref=badge_shield)
```

[...]

PR #288, opened 2019-06-21
Add dependabot endpoint

AndrewLeedham:
Based on this conversation: https://github.com/dependabot/feedback/issues/6#issuecomment-503994253

### identifier
Accepts a private repo id, for the example I filtered out `<repo_id>` since I don't have an actual private repo to add (and I don't think it is a good idea), and used greenkeeper because it is a repo I can't see ever implementing dependabot, so it shows both states.

fixes: #283

### DIFF
--- a/endpoints/dependabot.ts
+++ b/endpoints/dependabot.ts
@@ -9,20 +9,20 @@ import {
 export const meta: Meta = {
   title: 'Dependabot',
   examples: {
-    '/dependabot/github/dependabot/dependabot-core/?icon=dependabot': 'status',
-    '/dependabot/github/greenkeeperio/greenkeeper/<repo_id>?icon=dependabot': 'status (private repo)'
+    '/dependabot/dependabot/dependabot-core/?icon=dependabot': 'status',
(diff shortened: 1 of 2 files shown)


endpoints/dependabot.ts
```diff
@@ -0,0 +1,36 @@
...
+  BadgenServeMeta as Meta,
+  BadgenServeHandlers as Handlers,
+  BadgenServeHandlerArgs as Args
+} from '../libs/badgen-serve'
+
+export const meta: Meta = {
+  title: 'Dependabot',
+  examples: {
+    '/dependabot/dependabot/dependabot-core/?icon=dependabot': 'status',
+    '/dependabot/greenkeeperio/greenkeeper/<repo_id>?icon=dependabot': 'status (private repo)'
```
amio: `<repo_id>` in example might make user confused

AndrewLeedham: Fair point, addressed your concerns in the latest commit 😃

amio:
Awesome, thanks! @AndrewLeedham

tunnckoCore:
:tada:

Issue #290, opened 2019-06-24
All flat.badgen.net badges 500 error unless they have an icon!


AndrewLeedham:
Not sure what is causing it, but on https://flat.badgen.net/ all the images are broken apart from the select few that have a `?icon=` query. Could be a local issue to me, would be useful if someone can confirm they are seeing the same:

![image](https://user-images.githubusercontent.com/5557458/60020069-2b235d00-9687-11e9-8685-5222e73e448a.png)

amio:
🤦‍♂️ Rolled back to previous version

amio:
Fix deployed 🙏

amio:
Added uptime checker for flat.badgen.net, previously we only monitored badgen.net

![image](https://user-images.githubusercontent.com/215282/60058700-09c08080-971c-11e9-8b92-c33f1c2e1f7c.png)

PR #291, opened 2019-06-24
Resolve iconPromise to undefined by default

AndrewLeedham:
`.catch` [here](https://github.com/badgen/badgen.net/blob/master/libs/badgen-serve.ts#L61) only exists when a icon query exists, resolving the iconPromise to default allows it to catch regardless.
image

fixes: #290


amio:
>.<

Thanks! @AndrewLeedham

PR #295, opened 2019-06-29
Add codacy badges (coverage, grade)

xobotyi:
Codacy has no public API, so it has to parse SVG via regex;

### DIFF
--- a/endpoints/codacy.ts
+++ b/endpoints/codacy.ts
@@ -1,0 +1,96 @@
+import {
+  badgenServe,
+  BadgenServeHandlerArgs as Args,
+  BadgenServeHandlers as Handlers,
+  BadgenServeMeta as Meta
+} from '../libs/badgen-serve'
+// import { coverage as cov, coverageColor } from '../libs/utils'
+import got from '../libs/got'
+import { coverage as cov, coverageColor } from '../libs/utils'
+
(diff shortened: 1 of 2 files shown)


endpoints/codacy.ts
```diff
@@ -0,0 +1,96 @@
...
+  examples: {
+    '/codacy/c/f0875490cea1497a9eca9c25f3f7774e': 'coverage',
+    '/codacy/c/f0875490cea1497a9eca9c25f3f7774e/dev-master': 'branch coverage',
+    '/codacy/g/f0875490cea1497a9eca9c25f3f7774e': 'code quality',
+    '/codacy/g/f0875490cea1497a9eca9c25f3f7774e/dev-master': 'branch code quality'
+  }
+}
+
+export const handlers: Handlers = {
+  '/codacy/:type<c|g>/:projectId/:branch?': handler
```
amio: Let's use `coverage` and `grade` for type, it's meaningful and searchable,
and we can merge that two requests to one:

```
const svg = await got(`${uriBase}/${type}/${projectId}${query}`, { json: false })
```

xobotyi: Actually woke up with same thoughts =)
Reworked a bit now.

amio:
@xobotyi Thanks!

PR #296, opened 2019-07-01
Removes the useless horizontal scrollbar.

xobotyi:
It appears due to `100vw` but since there is a vertical scrollbar it has only `100vw - scrollbar width` of width.


amio:
Good call, thanks!

PR #298, opened 2019-07-15
Fix example for `label` override

loilo:
The example for the `label` parameter is currently the same as for `icon` (probably a copy&paste whoopsie). This PR makes it an actual example for `label` overriding.


amio:
@loilo Thanks!

PR #300, opened 2019-07-17
fix: ignore ParsedUrlQuery modification for serve

faustbrian:
Using `yarn dev` was not possible due to an error that was thrown while trying to assign the `icon` property to the `query` variable which only allows properties specified in `ParsedUrlQuery`.

```
72:9 Type 'string | undefined' is not assignable to type 'string | string[]'.
  Type 'undefined' is not assignable to type 'string | string[]'.
    70 |         params.status = simpleDecode(params.status)
    71 |
  > 72 |         query.icon = icon === '' ? params.subject : icon
       |         ^
    73 |
    74 |         if (query.style === undefined) {
    75 |           const host = req.headers['x-forwarded-host'] || req.headers.host

> Build error occurred
```

### DIFF
--- a/libs/badgen-serve.ts
+++ b/libs/badgen-serve.ts
@@ -69,6 +69,7 @@ export function badgenServe (handlers: BadgenServeHandlers): Function {
         params.subject = simpleDecode(params.subject)
         params.status = simpleDecode(params.status)
 
+        // @ts-ignore
         query.icon = icon === '' ? params.subject : icon
 
         if (query.style === undefined) {
--- a/libs/badgen-serve.ts
+++ b/libs/badgen-serve.ts
@@ -69,8 +69,9 @@ export function badgenServe (handlers: BadgenServeHandlers): Function {
         params.subject = simpleDecode(params.subject)
         params.status = simpleDecode(params.status)
 
-        // @ts-ignore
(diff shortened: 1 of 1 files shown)


styfle:
The problem is in `fetch-icon.ts` line 6 which returns `undefined`

https://github.com/badgen/badgen.net/blob/master/libs/fetch-icon.ts#L6

libs/badgen-serve.ts
```diff
@@ -69,6 +69,7 @@ export function badgenServe (handlers: BadgenServeHandlers): Function {
         params.subject = simpleDecode(params.subject)
         params.status = simpleDecode(params.status)
 
+        // @ts-ignore
         query.icon = icon === '' ? params.subject : icon
```
amio: Better to handle this case:

```javascript
        if (icon !== undefined) {
          query.icon = icon === '' ? params.subject : icon
        }
```

amio:
@styfle it was meant to be `undefined` when icon fetching failed, so issues with icon won’t break badge’s rendering.

Just realize we missing `@types/node` in `devDependencies`, I was using Node.js v10 definitions in my local dev :P

amio:
@faustbrian Thanks!

styfle:
Nice catch!

PR #303, opened 2019-08-08
Add support for Libraries.io dependency status

craigthomas:
This PR adds Libraries.io support to Badgen. In particular, it will generate a badge that lists the dependency status for a Github repository. Dependencies for the repository are examined from the Libraries.io API endpoint. Badges are generated on the following conditions:

* If the repository uses both deprecated and outdated dependencies, then a red `x deprecated, y outdated` badge is displayed.
* If the repository uses only deprecated dependencies, then a red `x deprecated` badge is displayed.
* If the repository uses only outdated dependencies, then an orange `y outdated` badge is displayed.
* If the repository uses dependencies that are up to date, then a green `up to date` badge is displayed.

### DIFF
--- a/endpoints/librariesio.ts
+++ b/endpoints/librariesio.ts
@@ -1,0 +1,58 @@
+import {
+  badgenServe,
+  BadgenServeHandlerArgs as Args,
+  BadgenServeHandlers as Handlers,
+  BadgenServeMeta as Meta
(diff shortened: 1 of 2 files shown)


craigthomas:
Sorry, wasn't fully ready for a pull request!

PR #312, opened 2019-09-30
fix: correct typo in codeclimate badge

wildlifehexagon:


amio:
Thanks! @wildlifehexagon

PR #317, opened 2019-10-29
Esnext badge

xobotyi:
Badge indicating if ESnext version bundled;

It is not tested due to inability to run build (it fails out of the box)


endpoints/npm.ts
```diff
@@ -96,6 +100,15 @@ async function unpkg (topic, pkg, tag = 'latest') {
         color: 'green'
       }
     }
+    case 'esnext': {
+      const bundled = typeof meta.esnext === 'string'
+
+      return {
+        subject: 'esnext',
+        status: `${bundled ? '' : 'not '} bundled`,
```
amio: According to this https://2ality.com/2017/06/pkg-esnext.html#package-authors

`esnext` field is for "untranspiled source code", maybe we should change "bundled" to "included" or something.

xobotyi: Mybe so. Not sure TBH.
As for me - `bundled` is more suitable here, but if you think `included` is better - i'll change it.

xobotyi:
@amio for now it is not so widespread as i'd like to, but things are starting to move and that field already described in [SchemaStore](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/package.json) and inherited from there to JetBrain's and Microsofts IDEs

endpoints/npm.ts
amio: I think we should be consistent with specification (if there is one 😅)

amio:
@xobotyi `esnext` field seems still in an unclear state, "untranspiled code" could be JavaScript, TypeScript, or something else. I think we better wait until it had a clear definition / be more widespread.

For your use case, a static badge should serve the purpose too I guess? It's not dynamically changing like a version badge, and you are free to use any words that fits the situation like "bundled". Then why not? 😀

xobotyi:
@amio welp, yes i can make it static for now. Just thought badgen can become like a broker that can ensure user that lib exports esnext version.
About what is ment under untranspiled version - it is coded in the name: **es**next, EcmaScript is a specification of JS, TypeScript is not an EcmaScript, its superset of ecma.

Anyways if you think it's not a time we can close the PR for now.

amio:
@xobotyi I think we should wait for the `esnext` field has broad usage and more solid definition/consense. Thanks for the PR anyway 🙏

PR #320, opened 2019-11-17
chore: migrate to Now v2

amio:
- new `create-badgen-handler`
- migrate `/endpoints` folder to `/api`
- update `now.json` to v2

Preview https://badgennet-amio.badgen.now.sh/


now.json
```diff
@@ -1,10 +1,26 @@
...
-  "version": 1,
-  "type": "docker",
+  "version": 2,
   "alias": [
-    "badgen.net",
-    "flat.badgen.net"
+    "v2.badgen.net"
   ],
+  "regions": ["sfo", "lhr", "hnd"],
+  "builds": [
```
styfle: You can delete the `builds` property and make sure your package.json `build` script outputs a `public` directory.

tsconfig.json
```diff
@@ -20,7 +20,7 @@
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "noEmit": false
+    "noEmit": true
```
styfle: Why noEmit?

now.json
```diff
@@ -1,10 +1,26 @@
...
+  ],
+  "env": {
+    "GH_TOKENS": "@badgen-gh-tokens",
+    "SENTRY_DSN": "@badgen-sentry-dsn",
+    "TRACKING_GA": "@badgen-tracking-ga"
+  },
+  "github": {
+    "autoAlias": false
+  },
   "files": [
```
styfle: I don't believe `files` property is supported for v2. You'll need a `.nowignore` file instead.

tsconfig.json
amio: It's not meant to be commited :P

now.json
amio: Nice, cleaner `now.json` ❤️

tsconfig.json
amio: `noEmit` is `false` by default, but if this line is removed, `npm run build` will add a `"noEmit": true` automatically 🤔:

```
The following changes are being made to your tsconfig.json file:
  - compilerOptions.noEmit to be suggested value: true (this can be changed)
```

amio:
@styfle Seems this was built multiple times, is it a bug or configuration issue?

image

https://zeit.co/badgen/badgennet/llnx6yz22

now.json
amio: @styfle Next.js doesn't allow using `public` for build output 🤔

https://github.com/zeit/next.js/blob/master/errors/can-not-output-to-public.md

styfle: Oh, I didn't realize you're using Next.js since you didn't use `@now/next`.

In that case, just follow Next.js documentation, no need `builds` key or really any build scripts in `package.json`

tsconfig.json
styfle: I've never seen that one before. A google search says this is coming from `create-react-app` 🤔 
https://stackoverflow.com/q/53945131/266535

amio: I'll just set it to `"noEmit": false`

now.json
amio: Prefer to keep it explicit :D

tunnckoCore:
Great! <3

PR #321, opened 2019-11-20
Fix badgesize url

styfle:
Double slashes are redirected to a single slash so one solution here is to omit the double slash.


amio:
Thanks @styfle!

PR #322, opened 2019-11-21
fix: typo for depentents -> dependents

kevinjalbert:


amio:
@kevinjalbert Thanks!

PR #324, opened 2019-11-26
chore: update now.json to use all regions

styfle:
It doesn't look like there is any central database dependency so you could run badgen in all regions.

See comment from @XhmikosR here https://github.com/styfle/packagephobia/pull/470#issuecomment-558744451

### DIFF
--- a/now.json
+++ b/now.json
@@ -1,6 +1,6 @@
 {
   "version": 2,
-  "regions": ["sfo", "lhr", "hnd"],
+  "regions": ["all"],
   "builds": [
     { "src": "package.json", "use": "@now/static-build" },
     { "src": "api/*.ts", "use": "@now/node" }


styfle:
On second look, the right side shows badgen.net is still using v1?

I spot a header `x-now-instance | 3532497067` which is not shown on the left side.

@XhmikosR What is the different between the left and right waterfall? Just the date?

XhmikosR:
Yeah, AFAICT.

styfle:
@XhmikosR then the answer to why badgen was slow was that it was booting up a whole docker image on v1. Today it uses v2 since #320 so your waterfall should be much different.

amio:
@XhmikosR I think this chart shows significant performance improvement with moving to Now V2, it was landed a week ago, after (NOV25) & before (NOV17):

image

PS:
- NOV25 is clearly a cache hit, NOV17 should be too, normally a fresh request should be longer than 1s.
- The "cache-control" header should be
   ```
   public, max-age=60, s-maxage=10800, stale-while-revalidate=604800, stale-if-error=604800
   ```
   not sure why it's
   ```
   public, max-age=60, stale-if-error=604800
   ```
   instead. Will fix that. We expect majority requests should be served by CDN.

XhmikosR:
The thing is that @styfle switched to preconnect so it's not a fair comparison :)

amio:
@XhmikosR The "Wating" phase take most part of request:

image

XhmikosR:
Yeah, but since preconnect is used, the time needed for the badgen connection is added to the HTML IIRC :)

I'll make some more tests from my region (Greece so theoretically I should hit London POP) with the direct badge links and report back.

amio:
> It doesn't look like there is any central database dependency so you could run badgen in all regions.

@styfle The concern is about GitHub api rate limitation. I was hoping if two requests from the different edges were forwarded to the same origin, the second one would get a cached response. In this case, we could save some quota by use limit origins 😅

Not sure if that's the case. If not, we should deploy to all regions.

styfle:
@amio The edge regions are only used as a cache so if the edge region has never been hit, the request will run on the origin server. You're going to have the same amount of serverless invocations regardless of which origins you select. If we knew which regions the GitHub API uses, perhaps it might make sense to use the same Now regions. But consider all of the badges that are generated which don't even need the GitHub API. I still think all regions is best for Badgen.

amio:
@styfle Yep, let's unleash it to the whole world 🌍🌎🌏

XhmikosR:
I believe now everything is faster :)

https://gtmetrix.com/compare/cXutlqMA/bJY5cQYb/NSIx2HwV

(do note that the waterfall doesn't compare 3 views)

amio:
Thank you! @styfle @XhmikosR

PR #330, opened 2019-12-07
Update azure-pipelines.ts

deepdotnet:
Added azure build, release, and deployment versions. Support for private azure projects.

### DIFF
--- a/api/azure-pipelines.ts
+++ b/api/azure-pipelines.ts
@@ -5,16 +5,18 @@ import { isNullOrUndefined } from 'util'
 export default createBadgenHandler({
   title: 'Azure Piplines',
   examples: {
-    '/azure-pipelines/build/status/yarnpkg/yarn/Yarn Acceptance Tests': 'build',
-    '/azure-pipelines/build/status/yarnpkg/yarn/Yarn Acceptance Tests/azure-pipelines': 'build (branch)',
+    '/azure-pipelines/build/status/yarnpkg/yarn/1': 'build status',
+    '/azure-pipelines/build/status/yarnpkg/yarn/1/1.21-stable': 'build status (branch)',
+    '/azure-pipelines/build/version/yarnpkg/yarn/1': 'build version',
+    '/azure-pipelines/build/version/yarnpkg/yarn/1/1.21-stable': 'build version (branch)'
   },
   handlers: {
-    '/azure-pipelines/:org/:project/:definition/:branch?': buildStatus,
     '/azure-pipelines/build/status/:org/:project/:definition/:branch?': buildStatus,
     '/azure-pipelines/build/version/:org/:project/:definition/:branch?': buildVersion,
     '/azure-pipelines/build/test/:org/:project/:definition/:branch?': buildTestResult,
     '/azure-pipelines/release/version/:org/:project/:definition?': releaseVersion,
     '/azure-pipelines/deployment/version/:org/:project/:definition/:environment?': deployedReleaseVersion,
+    '/azure-pipelines/:org/:project/:definition/:branch?': handler,
   }
 })
 
@@ -48,7 +50,8 @@ const azureDevOpsApiResponse = async (org: string, project: string, path: string
 }
 
 async function getLatestBuild ({ org, project, definition, branch = 'master'}: PathArgs) {
-  return await azureDevOpsApiResponse(org, project, `build/latest/${definition}?api-version=${getApiVersion(true)}&branchName=${branch}`)
+  const build = await azureDevOpsApiResponse(org, project, `build/builds?api-version=${getApiVersion(false)}&branchName=refs/heads/${branch}&definitions=${definition}&$top=1`)
+  return build.value[0]
 }
 
 async function getLatestRelease ({ org, project, definition}: PathArgs) {
@@ -94,9 +97,9 @@ async function buildTestResult ({ org, project, definition, branch = 'master'}: 
   const notExecuted: {outcome: string, count: number} = runStatistics.find( (value: { outcome: string; }) => value.outcome === 'NotExecuted')
   const failed: {outcome: string, count: number} = runStatistics.find( (value: { outcome: string; }) => value.outcome === 'Failed')
   
-  const passedCount = !isNullOrUndefined(passed) ? passed.count : 0
-  const notExecutedCount = !isNullOrUndefined(notExecuted) ? notExecuted.count : 0
-  const failedCount =  !isNullOrUndefined(failed) ? failed.count : total - passedCount - notExecutedCount
+  const passedCount = passed?.count ?? 0
+  const notExecutedCount = notExecuted?.count ?? 0
+  const failedCount =  failed?.count ?? total - passedCount - notExecutedCount
   
   const status = total == passedCount ? 'succeeded' : total == failedCount ? 'failed' : 'partially succeeded'
   const color = colors[status]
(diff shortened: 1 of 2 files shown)


deepdotnet:
I think it's not breaking change. I have kept old URL pattern as well for backword compatibility.

tunnckoCore:
ah yea probably, im a bit sleepy

api/azure-pipelines.ts
```diff
@@ -1,42 +1,130 @@
...
   title: 'Azure Piplines',
   examples: {
-    '/azure-pipelines/yarnpkg/yarn/Yarn Acceptance Tests': 'build',
-    '/azure-pipelines/yarnpkg/yarn/Yarn Acceptance Tests/azure-pipelines': 'build (branch)',
+    '/azure-pipelines/build/status/yarnpkg/yarn/Yarn Acceptance Tests': 'build',
+    '/azure-pipelines/build/status/yarnpkg/yarn/Yarn Acceptance Tests/azure-pipelines': 'build (branch)',
   },
   handlers: {
-    '/azure-pipelines/:org/:project/:definition/:branch?': handler
+    '/azure-pipelines/:org/:project/:definition/:branch?': buildStatus,
```
amio: This better be placed last as a deprecated fallback

api/azure-pipelines.ts
```diff
@@ -1,42 +1,130 @@
...
-  const response = await got(`https://dev.azure.com/${org}/${project}/_apis/build/status/${definition}?branchName=${branch}`, { json: false })
-  const contentType = response.headers['content-type']
-
-  if (!contentType.includes('image/svg+xml')) {
-    return {
-      subject: 'Azure Pipelines',
-      status: 'unknown',
-      color: 'grey'
-    }
+const AZURE_DEVOPS_TOKEN = process.env.AZURE_DEVOPS_TOKEN
```
amio: Need to add this to [`now.json`](https://zeit.co/docs/configuration/#project/env):

```diff
	{
	  "env": {
	    "GH_TOKENS": "@badgen-gh-tokens",
	    "SENTRY_DSN": "@badgen-sentry-dsn",
	    "TRACKING_GA": "@badgen-tracking-ga",
+	    "AZURE_DEVOPS_TOKEN": "@badgen-azure-token"
	  }
	}
```

api/azure-pipelines.ts
```diff
@@ -1,42 +1,130 @@
...
 import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
+import { isNullOrUndefined } from 'util'
 
 export default createBadgenHandler({
   title: 'Azure Piplines',
   examples: {
-    '/azure-pipelines/yarnpkg/yarn/Yarn Acceptance Tests': 'build',
-    '/azure-pipelines/yarnpkg/yarn/Yarn Acceptance Tests/azure-pipelines': 'build (branch)',
+    '/azure-pipelines/build/status/yarnpkg/yarn/Yarn Acceptance Tests': 'build',
+    '/azure-pipelines/build/status/yarnpkg/yarn/Yarn Acceptance Tests/azure-pipelines': 'build (branch)',
```
amio: It would be nice to have more examples showcasing all these features 👍

api/azure-pipelines.ts
```diff
@@ -1,42 +1,130 @@
...
+  const testResult = await getTestResultByBuild({org, project, build: build.id, minLastUpdatedDate: build.startTime, maxLastUpdatedDate: build.finishTime})
+  const runStatistics = testResult.value[0].runStatistics
+  const total: number = testResult.value[0].totalTests
+  const passed: {outcome: string, count: number} = runStatistics.find( (value: { outcome: string; }) => value.outcome === 'Passed')
+  const notExecuted: {outcome: string, count: number} = runStatistics.find( (value: { outcome: string; }) => value.outcome === 'NotExecuted')
+  const failed: {outcome: string, count: number} = runStatistics.find( (value: { outcome: string; }) => value.outcome === 'Failed')
+  
+  const passedCount = !isNullOrUndefined(passed) ? passed.count : 0
+  const notExecutedCount = !isNullOrUndefined(notExecuted) ? notExecuted.count : 0
+  const failedCount =  !isNullOrUndefined(failed) ? failed.count : total - passedCount - notExecutedCount
```
amio: We are using typescript ^3.7 now, these can be simplified to:

```js
  const passedCount = passed?.count ?? 0
  const notExecutedCount = notExecuted?.count ?? 0
  const failedCount = failed?.count ?? total - passedCount - notExecutedCount
```

deepdotnet: Thanks for the suggestion. Implemented the same.

api/azure-pipelines.ts
deepdotnet: Added build version example. I don't have any Azure DevOps public project which I can use as an example. Need to figure out how to put examples.

api/azure-pipelines.ts
deepdotnet: Done

deepdotnet:
I have added the old handler back as new code doesn't work with the definition name but with ID only.

deepdotnet:
@amio what needs to be done for above failure?

amio:
I need to add a azure devops token to now secrets.

@deepdotnet Is this token required? I tried locally, it works without a token.

api/azure-pipelines.ts
amio: @deepdotnet Can we use `yarnpkg/yarn` for all examples?

api/azure-pipelines.ts
```diff
@@ -1,18 +1,136 @@
-import cheerio from 'cheerio'
```
amio: `cheerio` need to be imported

deepdotnet: The token is not mandatory. The token is required for private projects.

tunnckoCore:
> The token is not mandatory. The token is required for private projects.

He could add it, but it will be for _his_ private repos i think? I'm not familiar much with azure.

api/azure-pipelines.ts
deepdotnet: Yes, I tried, I didn't find any release info in API response. If we can add example UTs, ATs, and release in yarnpkg/yarn, I can add examples here.

deepdotnet:
> 
> 
> > The token is not mandatory. The token is required for private projects.
> 
> He could add it, but it will be for _his_ private repos i think? I'm not familiar much with azure.

Yes, Azure DevOps APIs require an access token for private projects, however public projects can be accessed without an access token. Certain APIs require an access token even for public projects. Having the capability to provide access token in badgen is useful for people looking for on-premise/their own deployment.

amio:
> Azure DevOps APIs require an access token for private projects

@deepdotnet I thought it's required for requesting azure api. If we want to support private project, the token should be provided by user, but it's not a good idea to let user put their token in badge url, so we don't support private project on badgen.net.

We need to remove the env config from `now.json` 😅

api/azure-pipelines.ts
amio: At least we would need a working example to test if the handler works as expect.

tunnckoCore:
> If we want to support private project, the token should be provided by user

Yea, exactly what I meant.

deepdotnet:
> 
> 
> > Azure DevOps APIs require an access token for private projects
> 
> @deepdotnet I thought it's required for requesting azure api. If we want to support private project, the token should be provided by user, but it's not a good idea to let user put their token in badge url, so we don't support private project on badgen.net.
> 
> We need to remove the env config from `now.json` 😅

For the same reason, I initially didn't add it :). Will remove it again.

deepdotnet:
Removed Azure token from now.

amio:
> For the same reason, I initially didn't add it :). Will remove it again.

My bad 😝

PR #333, opened 2019-12-11
Add Docker image size support

lukechilds:
Shows the size for a Docker image.

Defaults to the latest tag and amd64 architecture as they are most popular but a  custom tag or arch can optionally be specified with `/docker/size/:scope/:name/:tag?/:architecture?`.

Screenshot 2019-12-11 at 19 21 20

Notes:

- `/docker/size/library/ubuntu` is a little slow, I've re-used the ubuntu image as that's what's in the other library examples. The reason it's slow is it has to paginate through a lot of API responses to find the tag as there are a huge amount of tags on Docker Hub for Ubuntu.
- I'm manually calculating the size instead of using `millify` because `millify` calculates MB when what we actually want is MiB to be consistent with the Docker Hub UI.

### DIFF
--- a/api/docker.ts
+++ b/api/docker.ts
@@ -7,13 +7,13 @@ export default createBadgenHandler({
   examples: {
     '/docker/pulls/library/ubuntu': 'pulls (library)',
     '/docker/stars/library/ubuntu': 'stars (library)',
-    '/docker/size/library/ubuntu/latest': 'size (library)',
+    '/docker/size/library/ubuntu/latest/amd64': 'size (library)',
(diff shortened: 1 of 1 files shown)


amio:
Thanks @lukechilds !

lukechilds:
Thanks for the quick review @amio.

And thanks for badgen, it's such a great service!

If you send me a Bitcoin address or Lightning Network invoice I'd love to send you a tip for your efforts.

amio:
So glad to hear you like it @lukechilds
No plan to accept donation through Bitcoin yet, but thanks for your kindness, and your helps!

Issue #189, opened 2018-11-06
The cache: query param to force refresh of the cache


tunnckoCore:
It still stunning huge. I don't know is it me or is it Badgen, but mostly always nothing works. Neither F5, neither another browser, neither visiting the page in npm, neither Ctrl+R. I'm not sure if it's the caching for chrome too (i aggressively use much chrome flags). But still, it seems that that's not the problem, because even with another browser it does not refresh them.

I suggest to add a query param that always gives you fresh one, so i will be able to add it to my badge links.

Because currently i'm forced to visit the badge's link directly.

styfle:
My guess is Now CDN is caching the response but this can be controlled by badgen by setting cache-control headers.

https://zeit.co/blog/now-cdn

tunnckoCore:
I don't think it's the Now, but yea maybe.

amio:
Here's the cache header for live badge:
```
cache-control: public, max-age=60, stale-while-revalidate=604800, stale-if-error=604800, s-maxage=360
cf-cache-status: MISS
```

And AFAIK, we may encounter these cache/latency for a live badge to refresh:
- __Browser Cache__ Controlled by `max-age=60`, CTRL+R would bypass this;
- __CDN Cache__ Controlled by `s-maxage=360`, in seconds. (6 mins);
- __API Latency__ Some api may not always serve latest data. (e.g, jsDelivr is faster than unpkg.com for fetching package.json, but it's got a much more delay for latest published version than unpkg.com, so I switched back (a1bb0ce8263ed7d7791f967740656da4723b512d) to unpkg.com. But still, unpkg.com has some delay, even npm's cli `npm v pkg` has 1 minute delay or so for newly published version).

Overall, a new version may have up to 10min (by my estimate) delay from publishing to shown on the badge.

amio:
Hi @tunnckoCore do you have any advice on this? If no further action we should take, I'm going to close this issue :D

amio:
@tunnckoCore I had an idea that we could add a `?cache=30` to limit the cache age on cdn, but make it 1 second at least internally, for perf's sake.

tunnckoCore:
Make sense. But I currently am not sure if it still an issue. ;d

amio:
:D Then just hold this issue for a while. I'm kinda busy recently anyway. Let's revisit it later.

tunnckoCore:
[...]

PR #338, opened 2020-01-18
feat: add a cache option

amio:
- Set default cdn cache age to ~~24hrs~~ 6hrs (lower server load)
- Use `?cache=300` in query string to limit cache to 5min (`300`) at least

Related to #189, cc @tunnckoCore


components/home-intro.tsx
```diff
@@ -43,6 +43,11 @@ export default function HomeIntro ({ isFlat = false }) {
         Custom badge scale
         <a href='/badge/docker/v1.2.3/blue?icon=docker&scale=2'>e.g.</a>
       </li>
+      <li>
+        <code>cache</code>
+        Live badge are cached in cdn for 24hrs (86400), you may limit it to 5min (300) at least.
```
wopian: "you may limit it to a minimum of 5min (300)." would be a better wording imo

amio: Great! Updated in fca1d94

tunnckoCore:
Cool :)

PR #345, opened 2020-01-27
chore: use npm api for "latest" tag

### LINKED ISSUE #292: Fallback to the original npm registry (or jsdelivr) when unpkg fail
For couple of hours I got error 500. First noticed that a ton of badges are just broken on ton of repos. So then I tried to create new project (for which I use mine `charlike` scaffolder which is again using the unpkg registry too) it turns out that unpkg returns 500 for non-cached packages (which is still strange cuz it should have mine pacakges already cached, but probably some cache invalidation happen in recent months).

So, we need to handle this case and fallback to some more reliable source of truth. I bet on jsdelivr, cuz as we know the npm registry sucks a bit.

amio:
prefer official npm api as much as possible

resolve #292

### DIFF
--- a/api/npm.ts
+++ b/api/npm.ts
@@ -3,6 +3,11 @@ import got from '../libs/got'
 import { millify, version, versionColor } from '../libs/utils'
 import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
 
+// https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md
+// https://github.com/npm/registry/blob/master/docs/download-counts.md
+// https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md
+// https://unpkg.com/
+
 export default createBadgenHandler({
   title: 'npm',
   examples: {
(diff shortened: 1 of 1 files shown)


api/npm.ts
```diff
@@ -66,17 +67,27 @@ async function handler ({ topic, scope, pkg, tag }: PathArgs) {
...
+  const meta = await (tag === 'latest' && pkg[0] !== '@' ? npmMetadata(pkg) : pkgJson(pkg, tag))
 
   switch (topic) {
     case 'version': {
       return {
-        subject: `npm${tag === 'latest' ? '' : '@' + tag}`,
+        subject: tag === 'latest' ? 'npm' : `npm@${tag}`,
         status: version(meta.version),
-        color: tag === 'latest' ? versionColor(meta.version) : 'cyan'
+        color: tag === 'latest' ? versionColor(meta.version) : 'blue'
```
tunnckoCore: hm, why always blue if not `latest`? it was cool that it's a bit different.

tunnckoCore: I think that no matter the dist-tag it should follow the `versionColor`. But when it's on some dist-tag AND above v1 then to be cyan.

amio: Let's go with __no matter the dist-tag it should follow the versionColor__

amio: Non-latest tag will be shown on badge label:

[](https://badgen.net/npm/v/yarn/legacy)

amio: Updated in 1999c2a

tunnckoCore: Good.

But why just decided to switch from the cyan, in the first place? :thinking:

amio: `cyan` is used for some unstable/future tag (alpha, canary, rc, etc), not suitable for tag like `legacy`, so I changed it to a more general/meaningless `blue` 😋

tunnckoCore: ooh is there still such detection in the versionColor, cool then.

PR #346, opened 2020-01-28
chore: replace universal-analytics with measurement-protocol

amio:
[2.5KB (measurement-protocol)](https://bundlephobia.com/result?p=measurement-protocol@0.0.2) vs [673KB (universal-analytics)](https://bundlephobia.com/result?p=universal-analytics@0.4.20) 💪


styfle:
@amio Is this open source?

I tried visiting https://github.com/amio/measurement-protocol but I get a 404.

amio:
> @amio Is this open source?
> 
> I tried visiting [amio/measurement-protocol](https://github.com/amio/measurement-protocol) but I get a 404.

Not yet, I just made it worked with current use case, still need more work to meet the plan for v1. Should be out in a few days, with a readme at least :D

styfle:
Nice! I might be interested in using this instead of the default google analytics script.

I've also considered https://github.com/lukeed/ganalytics too but it will probably have the same issues as the default GA script.

amio:
@styfle Made it public https://github.com/amio/measurement-protocol 🚀

> I've also considered lukeed/ganalytics too but it will probably have the same issues as the default GA script.

What issue?

styfle:
> What issue?

The issue I have with GA is inaccurate metrics because of the large number of developers who use adblockers in their browser. It makes the traffic look much smaller than it actually is.

I think your `measurement-protocol` will be much more accurate and it even has error logging 🤩

PR #351, opened 2020-02-01
feat: checks to support context

### LINKED ISSUE #349: Github Checks endpoint currently supports only "combined"
Me again. :laughing: It's not like `status` endpoint which support both "combined" and single ones.

https://api.github.com/repos/tunnckoCore/opensource/commits/master/check-runs

image

So, 

```
# by check name / action name - check.name
# that's the job name (in github actions case)
https://badgen.net/github/checks/tunnckoCore/opensource/master/test
https://badgen.net/github/checks/tunnckoCore/opensource/master/publish
https://badgen.net/github/checks/yarnpkg/berry/master/validating


# or by app (name) created the check - check.app.slug
https://badgen.net/github/checks/tunnckoCore/opensource/master/github-actions
https://badgen.net/github/checks/badgen/badgen.net/checks-api-with-context/lgtm
```

Because for example we can have LGTM (which uses Checks API too) and a Github Actions CI, and to want the badge to show only the state of lgtm, or only the state of ci. Or even you will be able to get a badge only for the macos ci builds, like 

```
# per os
https://badgen.net/github/checks/tunnckoCore/opensource/master/macos
https://badgen.net/github/checks/tunnckoCore/opensource/master/ubuntu
https://badgen.net/github/checks/tunnckoCore/opensource/master/windows
```

or

```
# per nodejs version
https://badgen.net/github/checks/tunnckoCore/opensource/master/12.x
https://badgen.net/github/checks/tunnckoCore/opensource/master/10.x
https://badgen.net/github/checks/tunnckoCore/opensource/master/8.x
```

---

And one idea: i think we should expose an endpoint which combines `checks` and `status`, probably call it `build`? or `ci`? So we won't think which one to use and when...

```
https://badgen.net/github/ci/tunnckoCore/opensource/:ref/:context
```

tunnckoCore:
close #349


tunnckoCore:
@amio, as seen here, the thing is that with `includes()` you can manually pick what you need - weather its `macos` build, `linux` (ubuntu), or just the `test` job, or just the lint, or just the `12.x` builds of all OSes, or just the `10.x` builds and etc.

tunnckoCore:
@amio can we proceed this one?

amio:
> can we proceed this one?

@tunnckoCore Let's put the `toLowerCase` part aside, move on with rest.

tunnckoCore:
:partying_face: :tada:

PR #352, opened 2020-02-01
fix: use includes and toLowerCase in status endpoint

### LINKED ISSUE #350: Github Status endpoint uses `.startsWith` instead of `.includes`
I remember I made it intentionally to use `.includes`. Also, it should use `toLowerCase(context)`

tunnckoCore:
close #350

### DIFF
--- a/api/github.ts
+++ b/api/github.ts
@@ -132,7 +132,7 @@ async function status ({ owner, repo, ref = 'master', context }: PathArgs) {
   const resp = await restGithub(`repos/${owner}/${repo}/commits/${ref}/status`)
 
   let state = typeof context === 'string'
-    ? resp!.statuses.filter(st => st.context.includes(context.toLowerCase()))
+    ? resp!.statuses.filter(st => st.context.toLowerCase().includes(context.toLowerCase()))
     : resp!.state
 
   if (Array.isArray(state)) {
--- a/api/github.ts
+++ b/api/github.ts
@@ -132,7 +132,7 @@ async function status ({ owner, repo, ref = 'master', context }: PathArgs) {
   const resp = await restGithub(`repos/${owner}/${repo}/commits/${ref}/status`)
 
   let state = typeof context === 'string'
-    ? resp!.statuses.filter(st => st.context.startsWith(context))
+    ? resp!.statuses.filter(st => st.context.includes(context.toLowerCase()))
     : resp!.state
 
   if (Array.isArray(state)) {


api/github.ts
```diff
@@ -132,7 +132,7 @@ async function status ({ owner, repo, ref = 'master', context }: PathArgs) {
   const resp = await restGithub(`repos/${owner}/${repo}/commits/${ref}/status`)
 
   let state = typeof context === 'string'
-    ? resp!.statuses.filter(st => st.context.startsWith(context))
+    ? resp!.statuses.filter(st => st.context.includes(context.toLowerCase()))
```
amio: Is there guarantee that context will always be lowercase?

tunnckoCore: Sorry, what? The thing is... Ooooh, we should also `st.toLowerCase()`, got it.

Probably tomorrow will update it.

amio: If not, then I guess we better leave it as is 😉

Autocorrect might not always works as expected

tunnckoCore:
> Is there guarantee that context will always be lowercase?

nope, it can be `LGTM.com`.

The idea behind using `includes` in the beginning was that we can pass `ci` as context for showing both circleci and travisci.

As for the lowercase, yea we should lowercase both the incoming context from the user and from the api response.

amio:
> The idea behind using `includes` in the beginning was that we can pass `ci` as context for showing both circleci and travisci.

+1

> As for the lowercase, yea we should lowercase both the incoming context from the user and from the api response.

We should do as least magic as possible, unless we can be sure the magic works 100%. Autocorrect cases in both sides and compare them, might broken in some future cases, if it happens, we cannot simply deprecate it then.

tunnckoCore:
>  the magic works 100%.

It will. It's not that big magic. It's convenience. Because end users don't (should not) know what the api returns. They simply would expect to pass `lgtm` as context and that to work. Because they won't know that it's `LGTM.com`. Of course you can add all that in docs, but that's unnecessary. 

People love convenience, that's why we all give up our privacy.

I really don't see how it can broke something in future. That's a standard practice to lowercase internally to make checking better & easier. And

amio:
Isn't the `context` comes from here?

image

`LGTM analysis: JavaScript` and `now`. User will know it's `LGTM`

tunnckoCore:
Actually, yea, for the status endpoint. But that's not the case with the `checks` endpoint, where I allowed  `context` to be the app name, or the "check name" (which is the job name in Github Actions case).

I don't see any problem adding all this for safety.

amio:
Suppose there's a new service called __"Magic Interface"__, it's context name is __"MagicInterface: JavaScript"__, when someone use it with `circici`, `travisci`, then user want to create a `ci` badge to shows all ci jobs, but

1. They may not noticed that `MagicInterface` is counted for the result
2. Usually all checks are passed, so the badge would looks correct at first, until someday. This make it harder for user to realize the cause.

User may add more accurate keyword to match what they want, or may not (for `circleci` and `travisci` case).

The other problem is, after this incident, every time when user use this feature (combined status), they would learnt to check if there is a `MagicInterface` or anything alike. That's a burden I want to avoid.

tunnckoCore:
> Suppose there's a new service called "Magic Interface", it's context name is "MagicInterface: JavaScript", when someone use it with circici, travisci

I can think of _any_ such thing currently. Usually services are either similar to CodeCov (`codecov: patch`, `lgtm analysis`, `now`, `security/snyk`, `continuous-integration/appveyor/branch`) or a github bot (`Semantic Pull Requests`, `WIP`), or triggered by GH Action. So there won't be such case (like this practicular case with `ci`), or at least in the most of the time. As usually, we handle the most common cases.

Now, yea, which is a service, which is a bot, which is using Checks and etc, is a bit not clear but... yea. :D 

Okay, anyway, I can agree to remain with `.startsWith`, but we definitely should lowercase both sides (so to be case insensitive which can be frustrating).

amio:
> So there won't be such case (like this practicular case with ci), or at least in the most of the time.

You still need to think about theses cases to make that statement right? And you need to go over __all__ cases to be confident that the statement is 100% correct  (currently).

If the convenience comes with a price like this, and this convenience is not a must-have, I tend not to have it.

tunnckoCore:
> @amio: Suppose there's a new service called

Okay, I still don't understanding this example. :thinking:

api/github.ts
```diff
@@ -132,7 +132,7 @@ async function status ({ owner, repo, ref = 'master', context }: PathArgs) {
   const resp = await restGithub(`repos/${owner}/${repo}/commits/${ref}/status`)
 
   let state = typeof context === 'string'
-    ? resp!.statuses.filter(st => st.context.startsWith(context))
+    ? resp!.statuses.filter(st => st.context.toLowerCase().includes(context.toLowerCase()))
```
amio, line 135: I think we can go with `includes` but not `toLowerCase` ;)

```suggestion
    ? resp!.statuses.filter(st => st.context.includes(context))
```

tunnckoCore: I'm exactly the opposite, haha. Can agree on using `startsWith` but not removing the `toLowerCase` :rofl: 

Services and urls should never be case sensitive. It's how such things always worked.

It's like it's `tunnckoCore`, but GitHub won't forbid them to request `https://github.com/tunnckocore` lowercased, no matter do they know how it's styled or not. It just doesn't make sense. There're always exceptions, like npm, but we should not be an exception, I don't see any reason.

vladimyr:
This whole function can be further simplified: https://diffy.org/diff/ynxrjy9u5wlrrzbu6vrxdpldi

@tunnckoCore Easiest way to achieve what you want is to use case-insensitive regex.

tunnckoCore:
> Easiest

Easiest. We are not talking about the easiest. Using regex is basically the same thing, I don't see any value in using regex. And there's nothing wrong in `toLowerCase().includes()` it's explicit and clear.

vladimyr:
> > Easiest
> 
> ... I don't see any value in using regex. And there's nothing wrong in `toLowerCase().includes()` it's explicit and clear.

Of course, there is nothing wrong with the former approach but using regex has indeed some bonus points compared to it:
1. It is less verbose
2. It is consisted of fewer function calls making it faster
3. It conveys the message more clearly. Ok this is subjective but when you see `pattern.test(something)` idiom you immediately know what is happening there. OTOH reading chain of functions imposes a bigger cognitive load until you parse out an idea behind it.

Let me rephrase myself ~easiest~ most elegant  😉

tunnckoCore:
It's not elegant too :D I was going to mention exactly that it's more clear when you have functions with pretty great names btw.

2 is subjective too, I don't think regex is faster than 2 builtin functions.

But all that is just the next style and preferences debate tho :D

vladimyr:
> But all that is just the next style and preferences debate tho :D

True that. 😄 

> 2 is subjective too, I don't think regex is faster than 2 builtin functions.

If it is measurable than it should be measured: https://runkit.com/vladimyr/string-methods-vs-regex 😁

tunnckoCore:
> If it is measurable than it should be measured

True. But for me it always shows that string methods is faster, haha, Firefox 75 (openSUSE). For Chromium 81+ too.

vladimyr:
> True. But for me it always shows that string methods is faster, haha, Firefox 75 (openSUSE). For Chromium 81+ too.

Oddly enough I posted it even though I also noticed it proves me wrong, go figure. 😁 I try to avoid arguing with facts, sometimes 😝

wopian:
> Firefox 75 (openSUSE). For Chromium 81+ too.

For what its worth, Runkit doesn't run the code in your browser. It's executed in a sandboxed environment (in this case it's how fast Node 10.20.1 on their infrastructure can execute these 2 queries)

Node 10.20.1:
- "string methods x 2,281,772 ops/sec ±0.54% (96 runs sampled)"
- "regex x 1,390,043 ops/sec ±0.15% (97 runs sampled)"

Edge 84 (https://jsbench.me/30kayjqqj9/1)
- toLowercase - 27,915,972 ops/s ±1.17%
- regex - 5,721,565 ops/s ±0.95% (79.5% slower)

vladimyr:
> For what its worth, Runkit doesn't run the code in your browser. It's executed in a sandboxed environment (in this case it's how fast Node 10.20.1 on their infrastructure can execute these 2 queries)

Which is exactly why I used it. I wanted to compare string methods vs. regex in Node because that is where badgen.net runs. Browser stats are irrelevant except they prove that using string methods is more performant wherever you end up running your code 🙃

tunnckoCore:
> Oddly enough I posted it even though I also noticed it proves me wrong, go figure. grin I try to avoid arguing with facts, sometimes

Hihihihi... same here.

> It's executed in a sandboxed environment

Good point tho.

Anyway, I think we can agree that using the string methods is better for all reasons - clear code, expressive, perf.

api/github.ts
amio: > Services and urls should never be case sensitive

Agree! It's a strong point I missed 😝

amio:
> Services and urls should never be case sensitive

That's the point I missed, totally agree on this, thanks @tunnckoCore !

tunnckoCore:
Thanks too :)

After 1-2 months of backing to my inner roots... I'm going to dev mode again :)
Hooooly... what a crappy period it was.

amio:
@tunnckoCore  💪

Issue #357, opened 2020-02-19
Support GitHub milestone percentage


wopian:
Milestones are kinda tucked away on GitHub - behind `Issues -> Milestones` and would be nice to be able to list the statuses of on-going projects from READMEs.

# Example

`/github/milestones/<repoOwner>/<repo>/<milestoneID>/<milestoneName>`
`/github/milestones/chrislgarry/Apollo-11/1/Comanche055` produces:
![](https://flat.badgen.net/badge/Comanche055/93%25?color=green)

![image](https://user-images.githubusercontent.com/3440094/74804839-51160580-52d9-11ea-9c5c-081ae3f4b7e4.png)

# API

REST https://developer.github.com/v3/issues/milestones/
GraphQL https://developer.github.com/v4/object/milestone/

amio:
That's nice!

PR #358, opened 2020-02-20
feat(github): add milestone percentages

wopian:
Implements #357 

Division by 0 is handled and follows GitHub by treating no issues as 0%. Percentages are floored to the nearest integer (e.g 94%)

### DIFF
--- a/api/github.ts
+++ b/api/github.ts
@@ -2,7 +2,7 @@ import cheerio from 'cheerio'
 import distanceToNow from 'date-fns/formatDistanceToNow'
 
 import got from '../libs/got'
-import { version, millify } from '../libs/utils'
+import { version, millify, coverageColor } from '../libs/utils'
 import { createBadgenHandler, BadgenError, PathArgs } from '../libs/create-badgen-handler'
 
 export default createBadgenHandler({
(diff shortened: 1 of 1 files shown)


amio:
PS, @wopian what do you think about an actual bar? built with `█▉▊▋▌▍▎▏` and `\u0020`(spaces):

styfle:
The actual bar looks really strange. I don't think it would be clear what it means.

I think it would be best to show a color and a percentage like code coverage.

image

image

amio:
Plus, I forgot there's another rule, __the badge color should represent `status`, not other info__

wopian:
image

image

amio:
Sorry I didn't make it clear 😂

I agree with @styfle 's opinion that __"a milestone is similar to test coverage -- its all the issues you have to fix to ship the milestone"__, and with another thought supporting it too: __"the badge color should represent `status`, the percent, not the implicit onSchedule"__.

So let's keep your original implementation, @wopian sorry for the extra works 🙏

amio:
@wopian Can you revert last change? then we are good to go!

amio:
@wopian Thanks!

tunnckoCore:
Sweeet!

PR #375, opened 2020-04-18
Add .git to .dockerignore

SuperSandro2000:
This prevents cache busting while building the Docker Image when only some git data changes like doing a fetching or rebasing without actual changes.


amio:
Thanks! @SuperSandro2000

PR #376, opened 2020-04-18
Filter out empty GitHub tokens

SuperSandro2000:
I had specific the GitHub token like this ``GH_TOKENS=xxxxx,`` and 50% of the requests didn't use a token and failed because of that. To not run into this issue again by accident I filtered out empty results after the split.

### DIFF
--- a/api/github.ts
+++ b/api/github.ts
@@ -76,7 +76,7 @@ const pickGithubToken = () => {
     })
   }
 
-  const tokens = GH_TOKENS.split(',')
+  const tokens = GH_TOKENS.split(',').filter(function(el) {return el.length != 0})
   return tokens[Math.floor(Math.random() * tokens.length)]
 }
 


amio:
I'm using this for testing locally, it works fine by far:

```bash
env GH_TOKENS='xxxxxx' npm run dev:api
```

Not sure about the issue you are running into, are you using a single or multiple tokens? @SuperSandro2000

SuperSandro2000:
I put an extra ``,`` at the end by accident and just wanted to make sure no one else runs into this.
``
env GH_TOKENS='xxxxxx,' npm run dev:api
``

amio:
I think we don't need to bother this case, just remove that extra `,` then it's all good

SuperSandro2000:
How about we make add a comment how to correctly populate this env?

amio:
@SuperSandro2000 That's nice to have

amio:
Thanks @SuperSandro2000 , let's make the change in another PR

SuperSandro2000:
What do you mean with your last comment?

amio:
@SuperSandro2000 I mean "add a comment on how to correctly populate this env" in another PR

PR #386, opened 2020-05-15
badge: add opam service

vladimyr:
Scrape package information from https://opam.ocaml.org using `cheerio`.

This adds a new handler:
```
/opam/:topic/:pkg
```
where the topic can be one of:
- `v` (version)
- `dm` (monthly downloads)
- `license`

## Preview

image

### DIFF
--- a/api/opam.ts
+++ b/api/opam.ts
@@ -1,0 +1,77 @@
+import cheerio from 'cheerio'
+import got from '../libs/got'
+import { millify, version, versionColor } from '../libs/utils'
+import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
+
+const OPAM_REPO_URL = 'https://opam.ocaml.org/packages/'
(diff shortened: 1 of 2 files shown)


api/opam.ts
```diff
@@ -0,0 +1,77 @@
...
+const OPAM_REPO_URL = 'https://opam.ocaml.org/packages/'
+
+type PackageInfo = {
+  version: string,
+  license: string,
+  downloads: number
+}
+
+export default createBadgenHandler({
+  title: 'opam',
```
amio, line 15: We can give it a full title for better seo and search matching

```suggestion
  title: 'OCaml Package Manager',
```

vladimyr: Thx for fixing that (and missing EOF newline) for me :+1:

amio:
Thanks! @vladimyr

PR #394, opened 2020-05-20
Add Snyk support

### LINKED ISSUE #340: Support Snyk
It supports both regular and monorepos.

Basically that's their badge

```
https://snyk.io/test/github/badgen/badgen.net/badge.svg?targetFile=package.json
```

or for package from a monorepo

```
https://snyk.io/test/github/tunnckoCore/opensource/badge.svg?targetFile=@tunnckocore/utils/package.json
https://snyk.io/test/github/tunnckoCore/opensource/badge.svg?targetFile=packages/glob-cache/package.json
```

The sad thing is that it has only 200 tests per month, which isn't not on per repo basis but on per "target file" basis, which in case of a monorepo is almost nothing. Currently it counted 22 tests for me for mine single monorepo. In the pricing page it says it is "Unlimited tests for open source / public repos" but in the dashboard the range is to 200. So.. I guess we'll see. But anyway.

I'm still not sure if I woould use it at all. Instead I think to use Pnpm + lockfile-lint.

vladimyr:
Vulnerability scan results are parsed from official Snyk's badge using `cheerio`.

This adds a new handler:
```
/snyk/:user/:repo/:branch?
```
Additionally `targetFile` query param can be passed to the handler.

## Preview

image

This closes #340

### DIFF
--- a/api/snyk.ts
+++ b/api/snyk.ts
@@ -1,20 +1,20 @@
 import cheerio from 'cheerio'
 import got from '../libs/got'
-import { createBadgenHandler, PathArgs, Query } from '../libs/create-badgen-handler'
+import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
 
 export default createBadgenHandler({
   title: 'Snyk',
   examples: {
     '/snyk/badgen/badgen.net': 'vulnerability scan',
     '/snyk/babel/babel/6.x': 'vulnerability scan (branch)',
-    '/snyk/tunnckoCore/opensource?targetFile=@tunnckocore/utils/package.json': 'vulnerability scan (custom path)'
+    '/snyk/tunnckoCore/opensource/master/@tunnckocore%2Futils%2Fpackage.json': 'vulnerability scan (custom path)'
   },
   handlers: {
-    '/snyk/:user/:repo/:branch?': handler
+    '/snyk/:user/:repo/:branch?/:targetFile?': handler
   }
 })
 
-async function handler ({ user, repo, branch = 'master' }: PathArgs, { targetFile }: Query) {
+async function handler ({ user, repo, branch = 'master', targetFile }: PathArgs) {
   const badgeUrl = `https://snyk.io/test/github/${user}/${repo}/${branch}/badge.svg`
   const searchParams = new URLSearchParams()
   if (targetFile) searchParams.set('targetFile', targetFile)
@@ -24,15 +24,15 @@ async function handler ({ user, repo, branch = 'master' }: PathArgs, { targetFil
   const $color = $('g[mask] path')
     .filter((_, el) => el.attribs.d?.startsWith('M90'))
     .first()
-    
+
   const $subject = $('g text')
     .filter((_, el) => parseInt(el.attribs.x, 10) === 45)
     .first()
-  
+
   const $status = $('g text')
     .filter((_, el) => parseInt(el.attribs.x, 10) === 100)
     .first()
-  
+
   const subject = $subject.text().trim() || 'vulnerabilities'
   const status = $status.text().trim()
   const color = ($color.attr('fill')?.trim() || '').replace(/^#/, '')
--- a/api/snyk.ts
+++ b/api/snyk.ts
@@ -1,0 +1,49 @@
+import cheerio from 'cheerio'
+import got from '../libs/got'
+import { createBadgenHandler, PathArgs, Query } from '../libs/create-badgen-handler'
+
+export default createBadgenHandler({
+  title: 'Snyk',
+  examples: {
+    '/snyk/badgen/badgen.net': 'vulnerability scan',
+    '/snyk/babel/babel/6.x': 'vulnerability scan (branch)',
+    '/snyk/tunnckoCore/opensource?targetFile=@tunnckocore/utils/package.json': 'vulnerability scan (custom path)'
+  },
+  handlers: {
+    '/snyk/:user/:repo/:branch?': handler
+  }
+})
+
+async function handler ({ user, repo, branch = 'master' }: PathArgs, { targetFile }: Query) {
+  const badgeUrl = `https://snyk.io/test/github/${user}/${repo}/${branch}/badge.svg`
+  const searchParams = new URLSearchParams()
+  if (targetFile) searchParams.set('targetFile', targetFile)
+  const svg = await got(badgeUrl, { searchParams }).text()
+  const $ = cheerio.load(svg, { xmlMode: true })
+
+  const $color = $('g[mask] path')
+    .filter((_, el) => el.attribs.d?.startsWith('M90'))
+    .first()
+    
+  const $subject = $('g text')
+    .filter((_, el) => parseInt(el.attribs.x, 10) === 45)
+    .first()
+  
+  const $status = $('g text')
+    .filter((_, el) => parseInt(el.attribs.x, 10) === 100)
+    .first()
+  
+  const subject = $subject.text().trim() || 'vulnerabilities'
+  const status = $status.text().trim()
+  const color = ($color.attr('fill')?.trim() || '').replace(/^#/, '')
+
+  if (!status || !color) {
+    const context = [
+      `${user}/${repo}/${branch}`,
+      targetFile && `targetFile=${targetFile}`
+    ].filter(Boolean).join(' ')
+    throw new Error(`Unknown Synk status: ${context}`)
+  }
+
+  return { subject, status, color }
+}
--- a/libs/badge-list.ts
+++ b/libs/badge-list.ts
@@ -41,6 +41,7 @@ export const liveBadgeList = [
   'codeclimate',
   'azure-pipelines',
   // quality & metrics
+  'snyk',
   'lgtm',
   'uptime-robot',
   'xo',
--- a/libs/create-badgen-handler.ts
+++ b/libs/create-badgen-handler.ts
@@ -11,10 +11,9 @@ import sentry from './sentry'
 import { BadgenParams } from './types'
 
 export type PathArgs = NonNullable<ReturnType<typeof matchRoute>>
-export type Query = NonNullable<Record<string, string | undefined>>
 
 export interface BadgeMaker {
-  (pathArgs: PathArgs, query: Query) : Promise<BadgenParams | undefined>;
+  (pathArgs: PathArgs) : Promise<BadgenParams | undefined>;
 }
 
 export interface BadgenServeConfig {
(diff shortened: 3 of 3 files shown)


api/snyk.ts
```diff
@@ -0,0 +1,49 @@
+import cheerio from 'cheerio'
+import got from '../libs/got'
+import { createBadgenHandler, PathArgs, Query } from '../libs/create-badgen-handler'
+
+export default createBadgenHandler({
+  title: 'Snyk',
+  examples: {
+    '/snyk/badgen/badgen.net': 'vulnerability scan',
+    '/snyk/babel/babel/6.x': 'vulnerability scan (branch)',
+    '/snyk/tunnckoCore/opensource?targetFile=@tunnckocore/utils/package.json': 'vulnerability scan (custom path)'
```
amio, line 10: We need a better schema for this, to keep the convention of "path args for content, query args for style"

vladimyr: How about converting the `targetFile` to matrix param? 🤔 https://www.logicbig.com/quick-info/web/matrix-param.html

amio: Great idea! but still feels a bit tricky to most people didn't know about matrix param before

amio: Maybe we need to choose between `branch` and `targetFile` then 😂

Which one would be most used? We can add the other one later as a nicer solution comes out.

vladimyr: > Maybe we need to choose between `branch` and `targetFile` then 😂
> 
> Which one would be most used? We can add the other one later as a nicer solution comes out.

That is a really hard question. It is quite often that projects have a dev branch set as the main one and periodically push stable code dumps to master/release. Now it is a question whether they want to use the badge as dev indicator on in-flux dev branch or as a message to end-users as part of their stable release. Maybe they want both making branch altering quite necessary. OTOH if they have monorepo they must have the ability to target partial packages by pointing to appropriate manifest/s. I can't really say which would be more useful. But if one must go it probably makes sense to let go `targetFile` because as an existing badge user I wouldn't expect that to be implemented 🙃

amio: Or we can support `targetFile` in additional path:
- `/snyk/:user/:repo/:branch?`
- `/snyk/:user/:repo/:branch/:targetFile+`

to monorepo user (needs `targetFile`), still better than nothing. to none-monorepo user (the majority), it's the best.

vladimyr: It is a nice compromise but there is a nasty edge case when git branch contains slash. In that case, you can't really safely assume where branch name ends and pathname starts. 🙄

amio: 🤦‍♂️

amio: Technically, we can use `%2F` as slash in branch section

vladimyr: Yeah, that would work! Let's go with that then. I'll update PR accordingly.

vladimyr:
Updated preview with `targetFile` as optional path argument:

image

amio:
Thanks! @vladimyr

tunnckoCore:
A bit late on that... but.

> Technically, we can use `%2F` as forward slash in branch section

Naah... that's pretty ugly. Is query param still supported? I don't want to mess with `.replace` (I have the path from pkg.repository.directory) on my templates generation.

vladimyr:
> Also, noticing that `@packages%2Fglob-cache%2Fpackage.json` doesn't work? And actually, any other that are in other dirs.

I'm not sure what do you mean by that? 😕 
https://badgen.net/snyk/tunnckoCore/opensource/master/@packages%2Fglob-cache%2Fpackage.json works just fine 

> Naah... that's pretty ugly. Is query param still supported?

It definitely isn't the nicest solution but it got implemented to keep the semantic separation between path and query arguments. `targetFile` isn't available as query param anymore.

tunnckoCore:
Oops. Probably I missed putting the branch name, haha.

> targetFile isn't available as query param anymore.

why? let's support it. And please add it to the examples, and use another package than `utils` ;p

vladimyr:
> why? let's support it.

I've removed upon @amio's suggestion https://github.com/badgen/badgen.net/pull/394#discussion_r429288568 so it probably isn't coming back.

> And please add it to the examples,

Add what exactly? 😕 

> and use another package than `utils` ;p

I simply took your example without giving it second thought. Sorry about unwanted publicity, I'll change that right away.

vladimyr:
@tunnckoCore I went with rollup/plugins as a sort of safe. However, if you have any wishes regarding promoting your own or work of others feel free to express them and I'll update PR accordingly 😉

tunnckoCore:
Yup okay with both. I meant to not be the `@tunnckocore/utils` specifically because it's kind of nothing that important. And if it points to some of mine it would be better to be another package of them :)

> I've removed it upon @amio's suggestion #394 (comment) so it probably isn't coming back.

@amio, it would be good to have both supported. :)

api/snyk.ts
```diff
@@ -0,0 +1,49 @@
...
+  },
+  handlers: {
+    '/snyk/:user/:repo/:branch?/:targetFile?': handler
+  }
+})
+
+async function handler ({ user, repo, branch = 'master', targetFile }: PathArgs) {
+  const badgeUrl = `https://snyk.io/test/github/${user}/${repo}/${branch}/badge.svg`
+  const searchParams = new URLSearchParams()
+  if (targetFile) searchParams.set('targetFile', targetFile)
```
tunnckoCore, line 20: Correct me if I'm wrong, but that seems to be supported? :thinking:

vladimyr: That is setting query param for an outbound request. But actual value isn't coming from inbound request's query param but `targetFile` path argument instead.

amio:
@tunnckoCore  @vladimyr We can use `/:targetFile*` to make it catch all rest paths:

```js
handlers: {
   '/snyk/:user/:repo/:branch?/:targetFile*': handler
}
```

then the badge url can be composed as:

`https://badgen.net/snyk/user/repo/branch%2Fslashed/path/to/package.json`

Looks fine to 99% use cases I guess. We just need an example showcasing this, but I'm not sure we could easily find one 😂

vladimyr:
It crossed my mind too but I dismissed it because yes it is certainly prettier but it is also inconsistent and kinda magical. There are lots of badges that end with the `/branch` segment. Theoretically we could convert those to rest params so users could use branch name containing slashes without any modifications. However, they might assume it will work elsewhere while in reality it just works for the last path segment. Using both branch names with slashes and custom paths isn't that common and in that case, users should be aware that they need to fit it into path segment using URL encoding.

tunnckoCore:
@vladimyr true. 

@amio, still. 

The reverse is equally bad as the current one. That's why query param can come quite handy and most clean. It doesn't break correct assumptions and defaults (that it's a branch), and if you are in monorepo and need paths, just pass a param. Consistency is better than magic in any of the sides.

amio:
> It crossed my mind too but...

Same feeling :D

> Consistency is better than magic in any of the sides.

Yea, so we are choosing consistency between the two:
- Path is solely for generating badge content, query string is solely for determine how we serve a badge (style, cache control)
- Consistent rule for branch name with slash

I prefer the first one. And the first one gives another benefits:
- Keep `createBadgenHandler` interface simpler (or we would need to pass extra argument like `targetFile` to handler, maybe even the whole `request` object in the future)
- More easily integration with other badge generating method, like `runkit` or `https`, or more posibility.

I think it's worth.

> However, they might assume it will work elsewhere while in reality it just works for the last path segment.

We should strengthen this convention 🤔 "the rest params only works for last segment (and only when it's necessary)", as a JavaScript programmer I guess it's fine since we already used to `function (a, b, ...rest) ()`

---

I think we can reduce the possible confusion by clarifying the routes:

```javascript
handlers: {
   '/snyk/:user/:repo/:branch?': handler, // branch is optional, no rest-paths support
   '/snyk/:user/:repo/:branch/:targetFile+': handler // branch is required, rest-paths support for targetFile
}
```

vladimyr:
> > However, they might assume it will work elsewhere while in reality it just works for the last path segment.
> 
> We should strengthen this convention 🤔 "the rest params only works for last segment (and only when it's necessary)", as a JavaScript programmer I guess it's fine since we already used to `function (a, b, ...rest) {}`

That is true but keep in mind that JavaScript programmer can easily determine the number of function arguments just by looking at commas in between because strings are always quoted. Here we have a different situation where slash is also a delimiter and segments aren't wrapped or contained in any way so perceived and the actual number of arguments might differ.
My point here is that it vaguely resembles rest params but it still requires a deeper understanding and sharp eye to figure out what is really happening there. That being said I'm still not convinced we should enable it.

tunnckoCore:
> I think it's worth.

Yup.

> Path is solely for generating badge content, query string is solely for determine how we serve a badge (style, cache control)

Ah, yea. Very good. But this badge's case is specific, and really I don't think we should include `targetFile` in the path.

The biggest problem is that `branch` and `targetFile` can have multiple forward slashes. But generally, if we follow the most common sense and 90-99% of the time, and follow the conventions... then we see a) that branch names usually have just 1 or 2 forward slashes (I think Renovate and dependabot have more? hm); and b) `targetFile` will be used in most of the time for monorepos that most of the cases are with 2 slashes: `packages/foo/package.json`, and 3+ slashes in rare cases (like for example the https://github.com/bubble-dev/_ monorepo and similar, where you have something like "monorepos inside monorepo" inside the `packages/` dir... which is ugly and that's why I use separate top level "roots" like `@tunnckocore/` for scoped, `packages/` non scoped, `configs/` for tools configs packages).

So, having said that... ah, yea it will be magic, but it will be different and better magic than current one, haha. And it's the cost for sane & conventional-based defaults. Like... join branch and the ...rest params, count the slashes and consider the last 3 parts to be the targetFile (if there `>2` slashes it has "targetFile", if less, then it's a branch) and the previous ones to be the branch name. So they will be able to just type `/snyk/:user/:repo/docs/update-readme/for/v1.2.2/packages/foo/package.json`

vladimyr:
> Aaaah... I don't know. I can implement it, it's not that big magic actually.. in text form it's ugly and sounds a big magic, but it's not, haha... :D

Here you go:
```diff
diff --git a/libs/create-badgen-handler.ts b/libs/create-badgen-handler.ts
index 17b8ac9..86f8820 100644
--- a/libs/create-badgen-handler.ts
+++ b/libs/create-badgen-handler.ts
@@ -11,9 +11,10 @@ import sentry from './sentry'
 import { BadgenParams } from './types'
 
 export type PathArgs = NonNullable<ReturnType<typeof matchRoute>>
+export type Query = NonNullable<Record<string, string | undefined>>
 
 export interface BadgeMaker {
-  (pathArgs: PathArgs) : Promise<BadgenParams | undefined>;
+  (pathArgs: PathArgs, query: Query) : Promise<BadgenParams | undefined>;
 }
 
 export interface BadgenServeConfig {
@@ -75,7 +76,7 @@ export function createBadgenHandler (conf: BadgenServeConfig): BadgenHandler {
 
     // Serve badge
     try {
-      const badgeParamsPromise = conf.handlers[matchedScheme](matchedArgs || {})
+      const badgeParamsPromise = conf.handlers[matchedScheme](matchedArgs || {}, query || {})
 
       let iconPromise: Promise<string | undefined> = Promise.resolve(undefined)
       if (typeof query.icon === 'string') {

```

After applying that patch all badge handlers will receive the second param holding query string arguments including those used for styling.

Issue #362, opened 2020-03-19
Support Dynamic Badges - Discord


donavanbecker:
Like Shields.io did here - https://shields.io/category/chat

amio:
Looks like two more live badges to go, discourse and matrix.

donavanbecker:
Can you do discord too?

amio:
Nice to have, let's put it on the list. PR's welcome 🤗

SuperSandro2000:
I would like to additionally request mastodon.

donavanbecker:
[shields.io - discord](https://github.com/badges/shields/tree/fcf6678a127c9679b0d68284b860181c2580fe26/services/discord)

amio:
Many thanks to @vladimyr !

donavanbecker:
So awesome!!

PR #396, opened 2020-05-20
Add Mastodon/Pleroma support

vladimyr:
Fetch followers for a given user from the default (mastodon.social) or custom Mastodon/Pleroma instance.

This adds a new handler:
```
/mastodon/follow/:userId/:instance?
```

## Preview

image

As proposed here: https://github.com/badgen/badgen.net/issues/362#issuecomment-615880134

### DIFF
--- a/api/mastodon.ts
+++ b/api/mastodon.ts
@@ -1,0 +1,26 @@
+import got from '../libs/got'
+import { millify } from '../libs/utils'
+import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
+
+export default createBadgenHandler({
+  title: 'Mastodon/Pleroma',
+  examples: {
+    '/mastodon/follow/1': 'followers count (mastodon.social)',
+    '/mastodon/follow/1/mas.to': 'followers count (custom Mastodon instance)',
+    '/mastodon/follow/1/pleroma.site': 'followers count (custom Pleroma instance)',
+  },
+  handlers: {
+    '/mastodon/follow/:userId/:instance?': handler
+  }
+})
+
+async function handler ({ userId, instance = 'mastodon.social' }: PathArgs) {
+  const prefixUrl = `https://${instance}/api/v1/`
+  const info = await got(`accounts/${userId}`, { prefixUrl }).json<any>()
+
+  return {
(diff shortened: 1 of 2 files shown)


api/mastodon.ts
```diff
@@ -0,0 +1,26 @@
...
+    '/mastodon/follow/:userId/:instance?': handler
+  }
+})
+
+async function handler ({ userId, instance = 'mastodon.social' }: PathArgs) {
+  const prefixUrl = `https://${instance}/api/v1/`
+  const info = await got(`accounts/${userId}`, { prefixUrl }).json<any>()
+
+  return {
+    subject: `follow @${info.username}@${instance}`,
```
amio, line 22: Two `@` looks a bit duplicate, maybe remove the first one?
```suggestion
    subject: `follow ${info.username}@${instance}`,
```

api/mastodon.ts
```diff
@@ -0,0 +1,26 @@
+import got from '../libs/got'
+import { millify } from '../libs/utils'
+import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
+
+export default createBadgenHandler({
+  title: 'Mastodon/Pleroma',
+  examples: {
+    '/mastodon/follow/1': 'followers count (mastodon.social)',
+    '/mastodon/follow/1/mas.to': 'followers count (custom Mastodon instance)',
+    '/mastodon/follow/1/pleroma.site': 'followers count (custom Pleroma instance)',
```
amio, line 10: It would be nice to make the description shorter, avoid that overflow in screenshot :)

api/mastodon.ts
vladimyr: I agree but that is a convention in Mastodon world:



https://mastodon.social/@Gargron

api/mastodon.ts
vladimyr: image

Better? 👨‍🎨

amio: Looks nice!

api/mastodon.ts
amio: Yea, let's go with convention 👍

amio:
Thanks! @vladimyr

PR #402, opened 2020-05-25
Update snyk examples

vladimyr:
Changing Synk examples per @tunnckoCore's request https://github.com/badgen/badgen.net/pull/394#issuecomment-633729983


amio:
@vladimyr Thanks!

PR #403, opened 2020-05-25
Support other providers for LGTM badge

### LINKED ISSUE #356: Support bitbucket/gitlab in LGTM badges
Currently, there is support only for GitHub repositories. However, GitLab and BitBucket repos exist with the prefix of /gl/ and /b/ instead of GitHub's /g/ on LGTM's API, so this should be a customisation option.

Example bitbucket LGTM: https://lgtm.com/projects/b/wegtam/bitbucket-youtrack-broker/?mode=list
Example gitlab LGTM: https://lgtm.com/projects/gl/nekokatt/hikari/?mode=list

vladimyr:
This closes #356 

@nekokatt Are you ok with using your work as badge example?

### DIFF
--- a/api/lgtm.ts
+++ b/api/lgtm.ts
@@ -16,16 +16,22 @@ export default createBadgenHandler({
     '/lgtm/grade/g/apache/cloudstack/java': 'grade (java)',
     '/lgtm/grade/g/apache/cloudstack': 'grade (auto)',
     '/lgtm/grade/g/systemd/systemd': 'grade (auto)',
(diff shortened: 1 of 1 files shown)


vladimyr:
> I think support for `github|bitbucket|gitlab` would be valuable too.

Actually I've changed it to support `g|github|bitbucket|gitlab` exclusively because badgen doesn't support short names elsewhere and `g` is added just to keep compatibility with existing badges.

tunnckoCore:
True :)

nekokatt:
Go for it :-)

amio:
@vladimyr Thanks!

PR #406, opened 2020-05-30
badge(github, npm): refactor dependents badge

vladimyr:
Millify dependents count.

### Before

image

--- 

### After

image

### DIFF
--- a/api/github.ts
+++ b/api/github.ts
@@ -407,16 +407,12 @@ async function repoStats ({topic, owner, repo, ...restArgs}: PathArgs) {
 }
 
 function dependents (type: DependentsType) {
-  const accept = [
-    'text/html',
-    'application/xhtml+xml',
-    'application/xml'
-  ].join(',')
-  const headers = { accept }
-
   return async function ({ owner, repo }: PathArgs) {
-    const url = `https://github.com/${owner}/${repo}/network/dependents`
-    const html = await got(url, { headers }).text()
+    const html = await got(`https://github.com/${owner}/${repo}/network/dependents`, {
+      headers: {
+        accept: 'text/html,application/xhtml+xml,application/xml'
+      }
+    }).text()
     const $ = cheerio.load(html)
     const $depLink = $(`a[href$="?dependent_type=${type}"]`)
 
--- a/api/github.ts
+++ b/api/github.ts
@@ -6,6 +6,8 @@ import { restGithub, queryGithub } from '../libs/github'
 import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
 import { version, millify, coverageColor } from '../libs/utils'
 
+type DependentsType = 'REPOSITORY' | 'PACKAGE'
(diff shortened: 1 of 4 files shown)


vladimyr:
I have extracted common logic for parsing dependents link into separate utility module.

api/github.ts
```diff
@@ -404,29 +406,24 @@ async function repoStats ({topic, owner, repo, ...restArgs}: PathArgs) {
...
 }
 
-function dependents (type: string) {
+function dependents (type: DependentsType) {
+  const accept = [
+    'text/html',
+    'application/xhtml+xml',
+    'application/xml'
+  ].join(',')
+  const headers = { accept }
```
amio, line 415: IMO the original one has better readability

```suggestion
  const headers = {
    accept: 'text/html, application/xhtml+xml, application/xml'
  }
```

vladimyr:
> Thanks @vladimyr for working on refactoring!

You are welcome 😃 

> For the `parseDependents` method, it heavily depends on page structure, I think it just happen to have some similarities between github and npm, and might change at any time. 

Locating the target DOM elements depends on the page structure while parsing its content remains the same and you are _guaranteed_ to have a number written there.

> So I would prefer to keep the dirty work (cheerio part) in their separate realm, and hide them as much as possible in a `parseDependents` function.

I've split DOM traversal and actual number extraction from the DOM element contents. Former got inlined at callsite; later got unified, and extracted to helper library. So only dirty work I could hide is DOM traversal and that is two lines of code: load HTML and target element with cheerio selector. There is no point in abstracting that away. Only thing I could do is leave things as is by closing this one 🙃

api/github.ts
vladimyr: Except original didn't have spaces after commas but nvm I'll return it to its original form. That being said you didn't really sell me _string-~a~rrays_ as a better option but this is your playground so I'm playing by your rules 😄

amio: I guess this `accept` header wouldn't change on a foreseeable future, it's more like a static string. In this case, less logic (like `join` method) means less thinking while reading :D

amio:
There's some reason I against the lastest version of `parseDepencents()`:

1. It's name seems doesn't sum up what it does straightly:

```typescript
export default ($el: Cheerio) => {
  const count = parseInt($el.text().replace(/,/g, ''), 10)
  return count ? millify(count) : 'unknown'
}
```

especially in a context like:
```
status: parseDependents($depLink),
```

2. The function handled how we process the text, well the text format is determined outside the function. These are strongly coupled 2 steps, they belong to a single function. Unless there's a convention for not doing this, and this behavior (passing a cheerio element to a function to parse it's text) is not a convention.

3. We can tolerate some redundancy in exchange for readability. This redundancy of 2 lines is totally fine to me :D

amio:
PR archived due to inactivity. Feel free to leave comments if you would like to continue with this, thanks! @vladimyr

PR #410, opened 2020-06-28
fix(github): status might be "cancelled"

amio:
There's a status "cancelled" in github checks we missed.

- before https://badgen.net/github/checks/vercel/vercel/6c81a873384f62ecfc49639d344f3f0aeb3dd558
- after https://badgennet-665t3sr35.vercel.app/github/checks/vercel/vercel/6c81a873384f62ecfc49639d344f3f0aeb3dd558

cc @tunnckoCore @styfle

### DIFF
--- a/api/github.ts
+++ b/api/github.ts
@@ -80,13 +80,19 @@ const statesColor = {
 
 function combined (states: Array<any>, stateKey: string = 'state') {
   if (states.length === 0) return 'unknown'
+
   if (states.find(x => x[stateKey] === 'error')) return 'error'
   if (states.find(x => x[stateKey] === 'failure')) return 'failure'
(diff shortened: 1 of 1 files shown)


amio:
Wait, just noticed github treated "cancelled" job as failure 🤔

https://github.com/vercel/vercel/commit/6c81a873384f62ecfc49639d344f3f0aeb3dd558
image

styfle:
I wish cancelled wasn't treated as failed.

When you click the red X, it shows cancelled jobs as a gray exclamation mark instead of the red X when you look at the individual checks in the popover.

cancelled

amio:
We all want it this way, so let's do it, "cancel" shouldn't be treated as failure. 💪

PR #412, opened 2020-07-07
feat: allow customn npm and github endpoints

krohrsb:
- Allow specifying of custom endpoints for NPM and Github. This is to allow usage in a self-hosted enterprise environment.
- Support querying NPM for all package info (scoped + non latest tags). Support abbreviated metadata queries for a bit better performance.
- Support to disable use of unpkg for package lookups (for selfhosted/private NPM use) when `NPM_REGISTRY` is defined.
- Added info on all supported environment variables.

Note:

My custom registry does not support abbreviated NPM metadata unfortunately, so some large internal packages can still time-out. Trying to think of a solution for this.

Maybe either configurable request timeout? Or perhaps an internal node cache (even if it is somewhat short, a minute or two) to avoid undue load on the registry? -- Or both. Open to suggestions, could be another PR.

### DIFF
--- a/api/npm.ts
+++ b/api/npm.ts
@@ -68,9 +68,14 @@ async function handler ({ topic, scope, pkg, tag }: PathArgs) {
 }
 
 async function npmMetadata (pkg: string, ver = 'latest'): Promise<any> {
-  const host = process.env.NPM_REGISTRY || "https://registry.npmjs.org"
-  if (pkg[0] === "@" || ver !== 'latest') {
-    const meta = await got(`${host}/${pkg}`).json<any>()
+  const host = process.env.NPM_REGISTRY || 'https://registry.npmjs.org'
+  if (pkg.startsWith('@') || ver !== 'latest') {
+    const meta = await got(`${host}/${pkg}`, {
+      // support querying abbreviated metadata https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md
+      headers: {
+        accept: 'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*'
+      }
+    }).json<any>()
     if (meta["dist-tags"][ver]) {
       return meta.versions[meta["dist-tags"][ver]]
     }
@@ -91,7 +96,7 @@ async function info (topic: string, pkg: string, tag = 'latest') {
   // ver === 'latest', non-scoped package use npmMetadata (npm), all others use unpkg
   // optionally disable unpkg to request all info from NPM
   const meta = await(
-    process.env.DISABLE_UNPKG === 'true' || (tag === "latest" && pkg[0] !== "@")
+    process.env.NPM_REGISTRY || (tag === "latest" && pkg.startsWith('@'))
       ? npmMetadata(pkg, tag)
       : pkgJson(pkg, tag)
   )
--- a/api/npm.ts
+++ b/api/npm.ts
@@ -69,7 +69,8 @@ async function handler ({ topic, scope, pkg, tag }: PathArgs) {
 
 async function npmMetadata (pkg: string, ver = 'latest'): Promise<any> {
   // only works for ver === 'latest', none-scoped package
-  const endpoint = `https://registry.npmjs.org/${pkg}/${ver}`
+  const host = process.env.NPM_URL || 'https://registry.npmjs.org'
+  const endpoint = `${host}/${pkg}/${ver}`
   return got(endpoint).json<any>()
 }
 
@@ -80,7 +81,8 @@ async function pkgJson (pkg: string, tag = 'latest'): Promise<any> {
 }
 
 async function info (topic: string, pkg: string, tag = 'latest') {
-  const meta = await (tag === 'latest' && pkg[0] !== '@' ? npmMetadata(pkg) : pkgJson(pkg, tag))
+  // when a custom registry is defined, tag=latest or package is scoped use npmMetadata
+  const meta = await (process.env.NPM_URL || (tag === 'latest' && pkg[0] !== '@') ? npmMetadata(pkg) : pkgJson(pkg, tag))
 
   switch (topic) {
     case 'version': {
--- a/api/npm.ts
+++ b/api/npm.ts
@@ -69,17 +69,15 @@ async function handler ({ topic, scope, pkg, tag }: PathArgs) {
 
 async function npmMetadata (pkg: string, ver = 'latest'): Promise<any> {
   const host = process.env.NPM_REGISTRY || "https://registry.npmjs.org"
-  if (pkg[0] === "@") {
+  if (pkg[0] === "@" || ver !== 'latest') {
     const meta = await got(`${host}/${pkg}`).json<any>()
     if (meta["dist-tags"][ver]) {
       return meta.versions[meta["dist-tags"][ver]]
-    } else {
-      throw new BadgenError({ status: '404', color: 'grey', code: 404 })
     }
-  } else {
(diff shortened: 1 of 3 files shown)


api/npm.ts
```diff
@@ -80,7 +81,8 @@ async function pkgJson (pkg: string, tag = 'latest'): Promise<any> {
 }
 
 async function info (topic: string, pkg: string, tag = 'latest') {
-  const meta = await (tag === 'latest' && pkg[0] !== '@' ? npmMetadata(pkg) : pkgJson(pkg, tag))
+  // when a custom registry is defined, tag=latest or package is scoped use npmMetadata
+  const meta = await (process.env.NPM_URL || (tag === 'latest' && pkg[0] !== '@') ? npmMetadata(pkg) : pkgJson(pkg, tag))
```
amio, line 85: Will the private npm registry api support query tags other than `latest`?

libs/github.ts
```diff
@@ -9,7 +9,7 @@ export function restGithub<T = any>(path: string, preview = 'hellcat') {
     authorization: `token ${pickGithubToken()}`,
     accept: `application/vnd.github.${preview}-preview+json`
   }
-  const prefixUrl = 'https://api.github.com/'
+  const prefixUrl = process.env.GITHUB_URL || 'https://api.github.com/'
```
amio, line 12: I think `GITHUB_API` might be a better naming for this, and `GITHUB_API_GRAPHQL` for the other one 😄

amio:
@krohrsb Thanks for the PR!

I don't have a private GitHub deployment, do you have one to verify if it works smoothly?

krohrsb:
I do. Aside from all of the examples 404ing for GitHub, it works well 😁

api/npm.ts
krohrsb: I'm testing using Artifactory and yes it does. Though it's a bummer NPM (at least public) does not. 

I wonder if instead of using unpkg one could just query the root package and pull the version out of the `versions` object?

I'll experiment with some things.

krohrsb:
@amio Updated with new changes. Also updated the PR description to surface those changes.

krohrsb:
Also, was wondering (if you choose to accept this PR) if you wanted me to document these variables / self hosted usage somewhere?

api/npm.ts
amio: > if instead of using unpkg one could just query the root package and pull the version out of the versions object?

I tried this method, it would be a huge json response for some long running packages like `webpack`

amio:
@krohrsb It would be great to have this documented 👍

api/npm.ts
```diff
@@ -68,9 +68,19 @@ async function handler ({ topic, scope, pkg, tag }: PathArgs) {
...
 async function npmMetadata (pkg: string, ver = 'latest'): Promise<any> {
-  // only works for ver === 'latest', none-scoped package
-  const endpoint = `https://registry.npmjs.org/${pkg}/${ver}`
-  return got(endpoint).json<any>()
+  const host = process.env.NPM_REGISTRY || "https://registry.npmjs.org"
+  if (pkg[0] === "@") {
+    const meta = await got(`${host}/${pkg}`).json<any>()
+    if (meta["dist-tags"][ver]) {
+      return meta.versions[meta["dist-tags"][ver]]
+    } else {
```
tunnckoCore, line 76: don't need the `else`, sorry for the nitpick ;d

krohrsb:
> @krohrsb It would be great to have this documented 👍

It just wasn't clear to me where to put docs (if there is a place other than the code)

krohrsb:
Pushed a change after some testing. And yes, testing with webpack on our internal registry with the master json + dist-tags method does time out occasionally. Not sure I can get around that, especially if needing to use our internal registry. Though, for self-hosted use-cases, its likely an edge case (having a package that large).

api/npm.ts
```diff
@@ -69,17 +69,15 @@ async function handler ({ topic, scope, pkg, tag }: PathArgs) {
 
 async function npmMetadata (pkg: string, ver = 'latest'): Promise<any> {
   const host = process.env.NPM_REGISTRY || "https://registry.npmjs.org"
-  if (pkg[0] === "@") {
+  if (pkg[0] === "@" || ver !== 'latest') {
```
tunnckoCore, line 72: what about `.charAt(0)`? @amio thoughts?

tunnckoCore:
> @krohrsb It just wasn't clear to me where to put docs (if there is a place other than the code)

In the readme. Probably another heading (h2-3?) like "Self-hosted solution" with explanation and the env vars.

api/npm.ts
amio: Guess I might choose `.startsWith()`, but all fine to me :D

api/npm.ts
```diff
@@ -80,7 +88,13 @@ async function pkgJson (pkg: string, tag = 'latest'): Promise<any> {
 }
 
 async function info (topic: string, pkg: string, tag = 'latest') {
-  const meta = await (tag === 'latest' && pkg[0] !== '@' ? npmMetadata(pkg) : pkgJson(pkg, tag))
+  // ver === 'latest', non-scoped package use npmMetadata (npm), all others use unpkg
+  // optionally disable unpkg to request all info from NPM
+  const meta = await(
+    process.env.DISABLE_UNPKG === 'true' || (tag === "latest" && pkg[0] !== "@")
```
amio, line 94: `DISABLE_UNPKG` seems very specific, if it's not a must-have, I prefer not to have it

Maybe we could find a better solution for this later, or maybe just wait better npm registry api to erase this problem.

libs/github.ts
```diff
@@ -9,7 +9,7 @@ export function restGithub<T = any>(path: string, preview = 'hellcat') {
     authorization: `token ${pickGithubToken()}`,
     accept: `application/vnd.github.${preview}-preview+json`
   }
-  const prefixUrl = 'https://api.github.com/'
+  const prefixUrl = process.env.GITHUB_API || "https://api.github.com/"
```
amio, line 12: Prefer single quotes

```suggestion
  const prefixUrl = process.env.GITHUB_API || 'https://api.github.com/'
```

libs/github.ts
```diff
@@ -20,7 +20,9 @@ export function queryGithub<T = any>(query) {
     accept: 'application/vnd.github.hawkgirl-preview+json'
   }
   const json = { query }
-  return got.post('https://api.github.com/graphql', { json, headers }).json<T>()
+  const endpoint =
+    process.env.GITHUB_API_GRAPHQL || "https://api.github.com/graphql"
```
amio, line 24:
```suggestion
    process.env.GITHUB_API_GRAPHQL || 'https://api.github.com/graphql'
```

amio:
We need a new section for all env vars:

- `GITHUB_TOKENS` (refactored from `GH_TOKENS`) (required for github badges)
- `GITHUB_API `
- `GITHUB_API_GRAPHQL`
- `NPM_REGISTRY `
- `SENTRY_DSN`
- `TRACKING_GA`

Let's put it in README.md with a heading `## Environments`

api/npm.ts
krohrsb: Alternatively I could disable unpkg usage only when a custom npm registry is defined? 

I kind of need to disable it as the primary usage for self hosting is for referencing internal packages. Packages that unpkg won't know about.

krohrsb:
Updated again per review comments. Updated PR description with some details. Thanks all.

amio:
Great work, thanks! @krohrsb

PR #418, opened 2020-08-10
feat: Added badge endpoint for Jenkins jobs

deepak-bhardwaj-ps:
Added a new badge endpoint for Jenkins jobs.

### DIFF
--- a/api/jenkins.ts
+++ b/api/jenkins.ts
@@ -38,7 +38,7 @@ const shortEnglishHumanizer = humanizeDuration.humanizer({
 });
 
 const statusToColor = (status: string) => {
-  return status.toUpperCase() == "SUCCESS" ? "green" : "red";
+  return status.toUpperCase() === "SUCCESS" ? "green" : "red";
 };
 
 const brokenBuildsToColor = (count: number) => {
--- a/api/jenkins.ts
+++ b/api/jenkins.ts
@@ -1,106 +1,108 @@
-import got from '../libs/got'
-import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
-import humanizeDuration from 'humanize-duration'
+import got from "../libs/got";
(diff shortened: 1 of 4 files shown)


deepak-bhardwaj-ps:
@amio please can you look into this PR

amio:
Sorry for missing this @deepak-bhardwaj-ps

api/jenkins.ts
```diff
@@ -0,0 +1,108 @@
...
+  return hours < 2 ? "green" : hours < 6 ? "orange" : "red";
+};
+
+async function jenkinsLastBuild({ hostname, job }: PathArgs) {
+  const endpoint = `https://${hostname}/${job}/lastBuild/api/json?tree=result,timestamp,estimatedDuration`;
+  return await got(endpoint).json<any>();
+}
+
+async function jenkinsBuilds({ hostname, job }: PathArgs) {
+  const endpoint = `https://${hostname}/${job}/api/json?tree=builds[number,status,timestamp,id,result]`;
```
amio, line 57: Do we need to handle paging here? Would it return a quite large json if there's a lot builds?

api/jenkins.ts
```diff
@@ -0,0 +1,108 @@
...
+      h: () => "h",
+      m: () => "m",
+      s: () => "s",
+      ms: () => "ms",
+    },
+  },
+});
+
+const statusToColor = (status: string) => {
+  return status.toUpperCase() == "SUCCESS" ? "green" : "red";
```
amio, line 41:
```suggestion
  return status.toUpperCase() === "SUCCESS" ? "green" : "red";
```

api/jenkins.ts
deepak-bhardwaj-ps: We can; however, the default behaviour of API restrict the response to 100 entries.

amio:
Nice work, thanks! @deepak-bhardwaj-ps

Issue #419, opened 2020-08-14
Add Badge for Github Dependabot


libinvarghese:
Since Github is supporting Dependabot [natively](https://docs.github.com/en/github/administering-a-repository/about-github-dependabot-version-updates), which is not compatible with the [Dependabot App](https://badgen.net/dependabot)

It would be great to have a badge to know if Dependabot is active on the repo.

eg: ![Dependabot](https://badgen.net/badge/Dependabot/Active/green)

amio:
Guess we may sunset the standalone dependabot badge at sometime.

PR #420, opened 2020-08-14
badge(github): add dependabot support

libinvarghese:
Since there is no API to get dependabot status, for now check if file `.github/dependabot.yml` exists.

fixes: #419


amio:
Thanks @libinvarghese !

PR #422, opened 2020-08-15
PR for adding some badges for Gitlab

### LINKED ISSUE #297: Is it possible to support commits and PR badges for GitLab?
Hello.

Is it possible to support more GitLab related badges? The same way it's done for GitHub, for example:

<div align='center'>

</div>

alexnaiman:
fixes #297 

I've added the following badges for Gitlab.

Screen Shot 2020-08-15 at 5 08 22 PM

I've covered some of them, but their API and documentation beat me many times 😆  (especially since I am not familiar with GraphQL or complex Gitlab project flows with lots of Jobs, tags and other things). Any suggestion and/or request for changing something is welcome 😄 

❗ When deploying the next time with this fix (if merged, of course) there is needed to add a new env variable named `GITLAB_TOKENS`(can be a list, same as GITHUB_TOKENS).  It should have `read_api` scope

### DIFF
--- a/api/gitlab.ts
+++ b/api/gitlab.ts
@@ -245,11 +245,7 @@ const makeRestCall = async ({ topic, owner, repo, ...restArgs }) => {
     'license': `/projects/${encodeURIComponent(`${owner}/${repo}`)}?license=true`,
   }
 
-  let restPath
-  switch (topic) {
-    default:
-      restPath = restPaths[topic]
-  }
+  let restPath = restPaths[topic]
 
   return restGitlab(restPath, fullResponsePaths.includes(topic))
 }
--- a/api/gitlab.ts
+++ b/api/gitlab.ts
@@ -1,6 +1,5 @@
 import millify from 'millify'
 import distanceToNow from 'date-fns/formatDistanceToNow'
-import got from '../libs/got'
 import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
 import { queryGitlab, restGitlab } from '../libs/gitlab'
 import { version } from '../libs/utils'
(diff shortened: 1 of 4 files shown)


amio:
Thanks @alexnaiman ! The long awaiting GitLab support 🎉
Will take a look on this soon

libs/gitlab.ts
```diff
@@ -0,0 +1,33 @@
...
+    authorization: `Bearer ${pickGitlabToken()}`,
+  }
+  const prefixUrl = process.env.GITLAB_API || 'https://gitlab.com/api/v4'
+  return fullResponse ? got.get(path, { prefixUrl, headers }) : got.get(path, { prefixUrl, headers }).json<T>()
+}
+
+function pickGitlabToken() {
+  const { GITLAB_TOKENS } = process.env
+  if (!GITLAB_TOKENS) {
+    throw new BadgenError({ status: 'token late required' })
```
amio, line 29: "late"?

api/gitlab.ts
```diff
@@ -0,0 +1,269 @@
...
+    'releases': `/projects/${encodeURIComponent(`${owner}/${repo}`)}/releases`,
+    'release': `/projects/${encodeURIComponent(`${owner}/${repo}`)}/releases`,
+    'license': `/projects/${encodeURIComponent(`${owner}/${repo}`)}?license=true`,
+  }
+
+  let restPath
+  switch (topic) {
+    default:
+      restPath = restPaths[topic]
+  }
```
amio, line 252: Seems this `switch` is not needed anymore

libs/gitlab.ts
alexnaiman: For some reason nodemon got stuck and it would not restart. That's how I tested, and I forgot about it. I'll remove it 😁

api/gitlab.ts
alexnaiman: Yup. I'll remove that too

libs/gitlab.ts
alexnaiman: @amio I changed the message

alexnaiman:
@amio 
> Thanks @alexnaiman
> 
> Here we need a `"GITLAB_TOKENS": "@badgen-gl-tokens"` in `now.json` as well

I added what you mentioned in the file

amio:
Thanks @alexnaiman !

PR #424, opened 2020-08-20
fix: now deploy command

amio:
Tested in https://github.com/badgen/badgen.net/runs/1006197731

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,4 +15,4 @@ jobs:
           node-version: 12.x
       - run: echo $GITHUB_SHA > static/sha
       - run: now --version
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -15,4 +15,4 @@ jobs:
           node-version: 12.x
       - run: echo $GITHUB_SHA > static/sha
       - run: now --version
-      - run: now -S badgen alias $(now -S badgen --token ${{ secrets.ZEIT_TOKEN }} -c) flat.badgen.net --token ${{ secrets.ZEIT_TOKEN }}
+      - run: now -S badgen alias $(now -S badgen --token ${{ secrets.VERCEL_TOKEN }} -c) flat.badgen.net --token ${{ secrets.VERCEL_TOKEN }}
```
styfle, line 18: You should use `vercel` instead 🙂

amio: I added this job to make flat.badgen.net a standalone deployment, other than sharing with badgen.net, since Vercel cache works at deployment level, not url level.

PR #428, opened 2020-09-06
Bundlephobia: dependency-count and tree-shaking badges

omgovich:
Hi! Love your project and use it quite often.

My hobby is creating tiny libraries and I know many people who need badges to promote that their JS-libraries have no dependencies (or have few of them). Also tagging a project as side-effect-free/tree-shakable would be very helpful for developers.

Example: https://github.com/pastelsky/bundlephobia/issues/224

So I decided to create a small PR to add these badges.
They get data from the bundlephobia's API that you already use, so it's actually not a big deal.
image

### DIFF
--- a/api/bundlephobia.ts
+++ b/api/bundlephobia.ts
@@ -59,7 +59,7 @@ async function handler ({ topic, scope, name }: PathArgs) {
       }
     case 'tree-shaking':
       return {
-        subject: 'tree-shaking',
+        subject: 'tree shaking',
         status: isTreeShakeable ? 'supported' : 'not supported',
         color: isTreeShakeable ? 'green' : 'red'
       }
--- a/api/bundlephobia.ts
+++ b/api/bundlephobia.ts
@@ -32,7 +32,11 @@ async function handler ({ topic, scope, name }: PathArgs) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


api/bundlephobia.ts
```diff
@@ -45,6 +51,18 @@ async function handler ({ topic, scope, name }: PathArgs) {
...
       }
+    case 'dependency-count':
+      return {
+        subject: 'dependency count',
+        status: dependencyCount,
+        color: 'blue'
+      }
+    case 'tree-shaking':
+      return {
+        subject: 'tree-shaking',
```
amio, line 62: Seems "tree shaking" is already a well known term https://en.wikipedia.org/wiki/Tree_shaking
We can call it `tree shaking` now :D
```suggestion
        subject: 'tree shaking',
```

amio (commented):
Looks great! Just a trivial change

api/bundlephobia.ts
omgovich: Yeah, you're right. Webpack docs use the same name:
https://webpack.js.org/guides/tree-shaking/

amio:
Thanks! @omgovich

PR #444, opened 2020-11-04
Delete CLI deployment and use Git

styfle:
You can add both domains to the same project because caching is handled per-domain now 👍


amio:
@styfle Awesome!

PR #453, opened 2020-12-13
chore: no need to run lint on pull requests

amio:
It already runs on push


wopian:
Push events don't happen on PRs from forked repos, so you won't get lint reporting from those

The following runs as a push event on non-fork PRs and as a pull_request event on fork PRs

```yaml
on:
  push:
  pull_request:
    branches:
    - '**:**'
```

amio:
@wopian Good point!

PR #460, opened 2020-12-17
Refactor OPAM badge

vladimyr:
Swapping cheerio powered SVG parsing with regex approach due to performance reasons.


amio:
@vladimyr Is there some performance issue with cheerio implementation?

I assuming cheerio approach might be slightly more stable than regex. But regex implementation looks quite neat!

vladimyr:
@amio I just wanted to speed things up, because regexes are still way faster than building DOM, and I also had a hunch that regex implementation would end up with a lower LOC count.

PR #468, opened 2020-12-28
Add Crystal shards service

vladimyr:
This adds a new handler powered by scraping https://shardbox.org:

```
/shards/:topic<v|license|crystal|dependents>/:shard
```
where the topic can be one of:
- `v` (version)
- `license`
- `crystal` (Crystal version)
- `dependents` (number of dependents)

## Preview

image

### DIFF
--- a/api/shards.ts
+++ b/api/shards.ts
@@ -1,0 +1,60 @@
+import got from '../libs/got'
+import { millify, version, versionColor } from '../libs/utils'
+import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
+
+const SHARDS_REPO_URL = 'https://shardbox.org/'
+
+const client = got.extend({ prefixUrl: SHARDS_REPO_URL })
+
+export default createBadgenHandler({
+  title: 'Crystal shards',
+  examples: {
+    '/shards/v/kemal': 'version',
+    '/shards/v/crinja': 'version',
+    '/shards/license/clear': 'license',
+    '/shards/crystal/amber': 'crystal version',
+    '/shards/dependents/lucky': 'dependents'
+  },
+  handlers: {
+    '/shards/:topic<v|license|crystal|dependents>/:shard': handler
+  }
+})
+
+async function handler({ topic, shard }: PathArgs) {
+  const html = await client.get(`shards/${shard}`).text()
+
+  switch (topic) {
+    case 'v':
+      const ver = html.match(/class="version">([^<]+)<\//i)?.[1].trim()
+      return {
+        subject: 'shards',
+        status: version(ver),
+        color: versionColor(ver)
+      }
+    case 'license': {
+      const license = html.match(/opensource.org\/licenses\/[^>]+?>([^<]+)<\//i)?.[1].trim()
+      return {
+        subject: 'license',
+        status: license || 'unknown',
+        color: 'blue'
+      }
+    }
+    case 'crystal': {
+      let ver = html.match(/Crystal<\/span>\s*<span[^>]*?>([^<]+)<\//i)?.[1].trim()
+      if (!ver || ver === 'none') ver = '*'
+      return {
+        subject: 'crystal',
+        status: version(ver),
+        color: 'green'
+      }
+    }
+    case 'dependents': {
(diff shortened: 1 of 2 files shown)


api/shards.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const SHARDS_REPO_URL = 'https://shardbox.org/'
+
+const client = got.extend({ prefixUrl: SHARDS_REPO_URL })
+
+export default createBadgenHandler({
+  title: 'Crystal shards',
+  examples: {
+    '/shards/v/kemal': 'version',
+    '/shards/v/crinja': 'version',
```
wopian, line 13: Is it necessary to have 2 examples of the same badge topic?

vladimyr: Absolutely not but it was modeled after https://badgen.net/npm :shrug:

amio: That's unnecessary, we need to update that 😅

amio:
@vladimyr I see a lot of regex works for parsing html, maybe we could make some common util function for most used cases? like

```typescript
extractInnerHTMLByClassName(className: string, tagName?: string)
extractInnerHTMLById(id:string, tagName?: string)
```

api/shards.ts
vladimyr: When I contributed my first badge I used `npm` as a sort of template. I always thought that the duplicated version badge is added intentionally to show that its color is calculated from the version itself. So I usually picked one pre v1 and one post v1 example to follow the suit. Reverting all that is out of scope for this PR but for starters, I'm gonna remove `crinja` because all (I guess :thinking:) shards are pre v1 anyway.

vladimyr:
> @vladimyr I see a lot of regex works for parsing html, maybe we could make some common util function for most used cases? like
> 
> ```ts
> extractInnerHTMLByClassName(htmlString: string, className: string, tagName?: string)
> extractInnerHTMLById(htmlString: string, id:string, tagName?: string)
> ```

I was tempted to try to implement something along those lines but I quickly figured out it simply isn't worth doing. Cases with simple CSS selectors are in fact outliars. Take this service for example. It scrapes data from https://shardbox.org shard page, lets say we go with https://shardbox.org/shards/kemal.

1. In order to retrieve **version** following regex is used:
	  ```js
	  /class="version">([^<]+)<\//
	  ```
	  and actual HTML (with preserved whitespace) looks like:
	  ```html
	        <span class="version">0.27.0</span>
	  ```
	  Hypothetical `extractInnerHTMLByClassName` could be used here but lets see other examples too... :+1: 

2. In order to retrieve **license** following regex is used:
	  ```js
	  /opensource.org\/licenses\/[^>]+?>([^<]+)<\//
	  ```
	  and, again, HTML source (with preserved whitespace) looks like:
	  ```html
	        <div class="metric">
	          <span class="metric-label">
	    <svg class="icon" width="24" height="24">
	      <use xlink:href="#icon-law"></use>
	    </svg>
	   License</span>
	          <span class="metric-value">
	            <a href="http://opensource.org/licenses/MIT" class="external">MIT</a>
	          </span>
	        </div>
	  ```
	  First, there isn't a good class name candidate here because both `metric-value` and `external` might and probably are used elsewhere too. So what it does it matches the tag that _contains_ `opensource.org/licenses/<license>` and extracts its content so no general helper could be used. :-1: 

3. Finally, in order to retrieve target **Crystal version** following regex is used:
	  ```js
	  /Crystal<\/span>\s*<span[^>]*?>([^<]+)<\//
	  ```
	  with given HTML source:
	  ```html
	      <div class="metric">
	        <span class="metric-label">
	    <svg class="icon" width="24" height="24">
	      <use xlink:href="#icon-crystal"></use>
	    </svg>
	   Crystal</span>
	        <span class="metric-value">0.35.0</span>
	      </div>
	    </div>
	  ```
	  Again, no good class name candidate so what it does - it matches the span that is preceded by (whitespace and) another span containing the word `Crystal` and extracts its content. And again it is a kind of _structure_ matching that no general helper could cover. :-1: 

Long story short regexes obviously are way more flexible but unfortunately, in most cases, we really need that flexibility as opposed to rolling our own generic helpers :upside_down_face:

amio:
I was thinking how can we prevent mistakes like #469, or make these regex better readability. But let's explore some other approach later. Thanks for the detailed explanation! @vladimyr

Issue #343, opened 2020-01-26
Move to another page generator


amio:
We are using Next.js static exports with now 2.0 deployments, now we want to limit function memory to a smaller size.

Adding a `functions` config to `now.json`:

```
  "functions": {
    "api/*.ts": {
      "memory": 512
    }
  },
```

caused this error:

```
> Error! The property `builds` is not allowed in now.json when using Now 2.0 – please remove it.
```

> Actually `builds` is allowed when `version` is `2`, just don't use it with `functions`

If we don't use `builds` (so no `@now/static` or `@now/static-build`), then we need to export pages to `public` dir (other than `dist`) to meet [zero-config](https://zeit.co/blog/zero-config)'s requirement.

While, Next.js doesn't allow [exporting to `public` dir](https://err.sh/zeit/next.js/can-not-output-to-public). There's a [workaround](https://github.com/zeit/now/issues/622#issuecomment-454482804) for this, but it need to setup a `builds` fields in `now.json`, which is disallowed when deploying to now v2, which is what happened in the beginning 🤦‍♂️

So finally, the solution could be moving to another static page generator, which exports to `public` folder.

styfle:
If you remove `builds` property from now.json, it should "just work" without any additional configuration.

amio:
> If you remove `builds` property from now.json, it should "just work" without any additional configuration.

Removed `builds` property https://badgennet-b7t1t7wa0.now.sh

@styfle htmls are fine but statics (js, css, etc) are missing

![image](https://user-images.githubusercontent.com/215282/73149096-af94fd00-40fa-11ea-86f4-bd870f10df8a.png)

styfle:
Seems like `static` should be `public` like the docs mention here: https://nextjs.org/docs/basic-features/static-file-serving

Also, you could probably move `/api` to `/pages/api` and then you can develop with `next dev` locally. It will likely speed up the production deployment time too since everything can be handed to Next.js instead of assuming each api needs separate dependencies.

PR #473, opened 2021-01-06
Change `@now/node` to `@vercel/node`

styfle:
> "@now/node" is deprecated and will stop receiving updates on December 31, 2020. Please use "@vercel/node" instead.

### Sources
- https://www.npmjs.com/package/@now/node
- https://vercel.com/blog/zeit-is-now-vercel

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -2,8 +2,8 @@
   "version": 2,
   "regions": ["all"],
   "builds": [
-    { "src": "package.json", "use": "@now/static-build" },
-    { "src": "api/*.ts", "use": "@now/node" }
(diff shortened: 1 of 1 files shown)


styfle:
I would also suggest removing `builds` as mentioned here: https://github.com/badgen/badgen.net/issues/343#issuecomment-579949478

amio:
Thanks! @styfle 

For removing `builds`, recently (after nextjs v10 👍) I had a try to migrate `/api/*` to nextjs's convention, but it didn't work well, I haven't dig it more but would give it another try later

styfle:
If you create a PR, I can take a look 👍

PR #486, opened 2021-02-01
Fix Perl release & runtime version

vladimyr:
Use API to get likes & runtime version - no more HTML scraping :tada: 
New _dependents_ badge :tada: 

## Before

image

---

## After

image


vladimyr:
In order to verify normalization output one can run following snippet:

```sh
perl -Mversion -e 'print version->parse("1.7044")->normal'
```

PR #487, opened 2021-02-01
Add Dart pub points & popularity badges

vladimyr:
Use pub API for retrieving likes, points & popularity scores.

## Before

image

---

## After

image


amio:
Thanks! @vladimyr

PR #490, opened 2021-02-06
Millify Github & npm dependents count

vladimyr:
This is a continuation of #406 with `cheerio` being replaced by regexes :tada: 

## Before

image

---

## After

image

### DIFF
--- a/api/github.ts
+++ b/api/github.ts
@@ -1,10 +1,11 @@
-import cheerio from 'cheerio'
 import distanceToNow from 'date-fns/formatDistanceToNow'
 
 import got from '../libs/got'
 import { restGithub, queryGithub } from '../libs/github'
 import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
-import { version, millify, coverageColor } from '../libs/utils'
+import { coverageColor, millify, version } from '../libs/utils'
+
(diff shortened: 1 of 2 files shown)


api/github.ts
```diff
@@ -454,29 +455,24 @@ async function repoStats ({topic, owner, repo, ...restArgs}: PathArgs) {
...
+        subject,
+        status: 'invalid',
+        color: 'grey'
+      }
+    }
     return {
-      subject: type === 'PACKAGE' ? 'pkg dependents' : 'repo dependents',
-      status: parseDependents(html, type),
+      subject,
+      status: millify(count),
```
amio, line 474: `count` might be a `NaN`, which would cause `millify` throw an error

vladimyr: True, I'll handle it explicitly.

api/npm.ts
```diff
@@ -163,24 +162,24 @@ const download = async (period: string, npmName: string, tag = 'latest') => {
   }
 }
 
-// https://github.com/astur/check-npm-dependents/blob/master/index.js
```
vladimyr, line 166: Not relevant anymore since I'm not scraping HTML but hydration data instead (JSON).

api/github.ts
amio: How I missed that block 🙈

PR #492, opened 2021-02-07
Fix Azure Pipeline badges

vladimyr:
Through refactor of Azure DevOps REST API client with strict URL param validation, non-hard coded default branch and easier build test results fetching using `test/ResultSummaryByBuild` endpoint. :surfer: 
Plus, the Azure badge is now being _parsed_ using regexes instead of `cheerio`.

## Before

image

---

## After

image


amio (approved):
This is great! Thanks @vladimyr

PR #494, opened 2021-02-07
Use `URLSearchParams#set` instead of `URLSearchParams#append`

vladimyr:
`append` is meant to be used for setting array params.


amio:
TIL :D

PR #500, opened 2021-02-08
Introduce Gitter and Matrix member count badges

vladimyr:
This PR adds support for Gitter and Matrix badges. Former is powered by scraping public Gitter rooms while latter uses Matrix client API to obtain public room information. :tada: 

## Preview

image

In response to https://github.com/badgen/badgen.net/issues/362#issuecomment-601659346
> Looks like two more live badges to go, discourse and matrix.

One down, one more to go :wink:

### DIFF
--- a/api/matrix.ts
+++ b/api/matrix.ts
@@ -8,8 +8,7 @@ const BRAND_COLOR = 'black'
 export default createBadgenHandler({
   title: 'Matrix',
   examples: {
-    '/matrix/members/%23rust/matrix.org': 'members',
-    '/matrix/members/Manjaro/matrix.org': 'members',
+    '/matrix/members/rust/matrix.org': 'members',
     '/matrix/members/thisweekinmatrix': 'members',
     '/matrix/members/archlinux/archlinux.org': 'members',
     '/matrix/members/redom_redom/gitter.im': 'members'
--- a/api/matrix.ts
+++ b/api/matrix.ts
@@ -1,0 +1,74 @@
+import got from '../libs/got'
+import { millify } from '../libs/utils'
+import { fetchMembersCount as fetchGitterMembersCount } from './gitter'
+import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
+
+const BRAND_COLOR = 'black'
+
+export default createBadgenHandler({
+  title: 'Matrix',
+  examples: {
+    '/matrix/members/%23rust/matrix.org': 'members',
+    '/matrix/members/Manjaro/matrix.org': 'members',
+    '/matrix/members/thisweekinmatrix': 'members',
+    '/matrix/members/archlinux/archlinux.org': 'members',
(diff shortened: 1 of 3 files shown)


amio:
@vladimyr You are awesome!

vladimyr:
Thank you! Although I'm just a regular guy having fun with your awesome creation!

api/matrix.ts
```diff
@@ -0,0 +1,74 @@
...
+import { millify } from '../libs/utils'
+import { fetchMembersCount as fetchGitterMembersCount } from './gitter'
+import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
+
+const BRAND_COLOR = 'black'
+
+export default createBadgenHandler({
+  title: 'Matrix',
+  examples: {
+    '/matrix/members/%23rust/matrix.org': 'members',
```
amio, line 11: Looks like this `%23` can be omitted

```suggestion
    '/matrix/members/rust/matrix.org': 'members',
```

api/matrix.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+async function findPublicRoom(roomName: string, server: string, homeserver?: string) {
+  homeserver = homeserver || await getHomeserver(server)
+  const roomAlias = `#${roomName}:${server}`
+  const endpoint = `${homeserver}/_matrix/client/api/v1/publicRooms`
+  const searchParams = new URLSearchParams({ limit: '500' })
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    const { chunk, next_batch } = await got(endpoint, { searchParams }).json<any>()
```
amio, line 62: Would be nice to give it a shorter timeout, for non-exists rooms. (default timeout is 1 min)

amio (commented):
There's a long timeout (1min) for non-exist rooms, would be nice to make it a quicker fail

api/matrix.ts
vladimyr: Yeah, I put it there intentionally so folks are aware it is an optional thing. Do we still want git rid of it?

vladimyr:
I'm not sure how to approach matrix timeout. If I abort scanning quicker we might miss a valid match :shrug:

api/matrix.ts
amio: I think people would always prefer the simpler form

amio:
@vladimyr Guess we can validate room with [alias api](https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-directory-room-roomalias) at first?

vladimyr:
Worth taking a shot, thx for the suggestion 👍

api/matrix.ts
vladimyr: Done :heavy_check_mark:

vladimyr:
@amio I believe we are good now. If the room alias can not be resolved, the handler fails immediately. :rocket:

amio (approved):
Thanks! @vladimyr

PR #504, opened 2021-02-15
Remove `cheerio`

vladimyr:
It can be safely removed since all handlers migrated away from using it. :tada:


amio:
One less dep 🚀

PR #507, opened 2021-02-16
Add support for Open VSX

idleberg:
Open VSX is an alternative to the Visual Studio Code Marketplace, operated by the Eclipse Foundation

### DIFF
--- a/api/open-vsx.ts
+++ b/api/open-vsx.ts
@@ -22,43 +22,38 @@ async function handler ({ topic, pkg, namespace }: PathArgs) {
 
   switch (topic) {
     case 'v':
-    case 'version': {
+    case 'version':
       return {
         subject: `version`,
         status: version(data.version),
         color: versionColor(data.version)
       }
-    }
     case 'd':
-    case 'downloads': {
+    case 'downloads':
       return {
         subject: 'downloads',
         status: millify(data.downloadCount),
         color: 'green'
       }
-    }
     case 'l':
-    case 'license': {
+    case 'license':
       return {
         subject: 'license',
         status: data.license || 'unknown',
         color: 'blue'
       }
-    }
     case 'r':
-    case 'reviews': {
+    case 'reviews':
       return {
         subject: 'reviews',
         status: millify(data.reviewCount),
         color: 'green'
       }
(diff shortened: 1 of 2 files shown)


libs/badge-list.ts
```diff
@@ -74,6 +74,7 @@ export const liveBadgeList = [
   'liberapay',
   'runkit',
   'https',
+  'open-vsx',
```
vladimyr, line 77: Suggestion: place it near other extension stores to keep the list organized :bulb:

api/open-vsx.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+})
+
+async function handler ({ topic, pkg, namespace }: PathArgs) {
+  const endpoint = `https://open-vsx.org/api/${namespace}/${pkg}`
+  const data = await got(endpoint).json<any>()
+
+  switch (topic) {
+    case 'v':
+    case 'version': {
```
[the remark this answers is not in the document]
idleberg: Hmm, copy&paste from the apm module 😉

api/open-vsx.ts
```diff
@@ -0,0 +1,59 @@
...
+export default createBadgenHandler({
+  title: 'Open VSX',
+  examples: {
+    '/open-vsx/v/idleberg/electron-builder': 'version',
+    '/open-vsx/r/idleberg/electron-builder': 'reviews',
+    '/open-vsx/l/idleberg/electron-builder': 'license',
+    '/open-vsx/d/idleberg/electron-builder': 'downloads'
+  },
+  handlers: {
+    '/open-vsx/:topic/:namespace/:pkg': handler
```
vladimyr, line 15: @amio Is it preferred to use `:topic<v|version|r|reviews|l|license|d|downloads>` or unconstrained `:topic` param in combination with default _unknown topic_ case?

amio: Both are fine, if there are lots of options, an unconstrained  `:topic` would be handy, if it's just a few, `:topic<a|few|option>` would be prefered. While there's no clear line between the two.

amio:
Thanks! @idleberg

PR #508, opened 2021-02-16
Pin Node version with Volta

idleberg:
This is my first of two proposals that makes working with a specific Node version easier. I separated them into two PRs, so you can decide if you want to use one, both, or none of the proposals.

[Volta](https://volta.sh/) is a fairly new tool to manage Node (and other) versions. Once installed globally, no action is required by the user to switch the Node version for a project. On the downside, pinning a Node version only works for the full SemVer range (so just specifying `"12"` won't work).


vladimyr:
FWIW we could bump node version up to `14.x` https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version :bulb:

amio:
Let's use node v14 👍

amio:
@idleberg Thanks!

PR #509, opened 2021-02-16
Add .nvmrc

idleberg:
This is my second of two proposals that makes working with a specific Node version easier. I separated them into two PRs, so you can decide if you want to use one, both, or none of the proposals.

[Node Version Manager](https://github.com/nvm-sh/nvm) is a widely used tool that allows installing multiple Node versions and switching between them easily. It supports "versionish" ranges, so it will always use the latest version of Node 12.x in this case. On the downside, it requires user action (`nvm use`) to activate, though I assume that it can be automated in editors such as VSCode.


vladimyr:
> It supports "versionish" ranges, so it will always use the latest version of Node 12.x in this case.

If you don't mind please verify it works with [`fnm`](https://fnm.vercel.app) too.

idleberg:
> If you don't mind please verify it works with fnm too.

Works for me (tested on macOS and Windows 10)

PR #512, opened 2021-02-21
Add support for Open VSX ratings

idleberg:
I somehow missed the ratings option in the Open VSX API. But now that there are two topics with the letter `r`, I'm having second thoughts about the abbreviated forms. Any opinions on whether I should remove that or do you fear a breaking change like this?

I'm wondering why there's no recommendation or at least a standardized way for some of these (e.g. I've seen `d` and `dl` as short forms for `downloads`). Personally, when in doubt, I'd opt for the full name topics.

### DIFF
--- a/api/open-vsx.ts
+++ b/api/open-vsx.ts
@@ -6,11 +6,11 @@ import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
 export default createBadgenHandler({
   title: 'Open VSX',
   examples: {
-    '/open-vsx/version/idleberg/electron-builder': 'version',
+    '/open-vsx/downloads/idleberg/electron-builder': 'downloads',
+    '/open-vsx/license/idleberg/electron-builder': 'license',
     '/open-vsx/rating/idleberg/electron-builder': 'rating',
     '/open-vsx/reviews/idleberg/electron-builder': 'reviews',
-    '/open-vsx/license/idleberg/electron-builder': 'license',
-    '/open-vsx/downloads/idleberg/electron-builder': 'downloads'
+    '/open-vsx/version/idleberg/electron-builder': 'version'
   },
   handlers: {
     '/open-vsx/:topic<downloads|license|rating|reviews|version>/:namespace/:pkg': handler
--- a/api/open-vsx.ts
+++ b/api/open-vsx.ts
@@ -55,7 +55,7 @@ async function handler ({ topic, pkg, namespace }: PathArgs) {
     default:
       return {
         subject: 'Open VSX',
-        status: 'unknown topic',
+        status: 'unknown',
         color: 'grey'
       }
   }
--- a/api/open-vsx.ts
+++ b/api/open-vsx.ts
@@ -13,7 +13,7 @@ export default createBadgenHandler({
     '/open-vsx/version/idleberg/electron-builder': 'version'
   },
   handlers: {
-    '/open-vsx/:topic<downloads|license|rating|reviews|version>/:namespace/:pkg': handler
+    '/open-vsx/:topic<d|l|license|rating|reviews|v|version>/:namespace/:pkg': handler
   }
 })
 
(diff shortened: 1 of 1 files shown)


amio:
> Personally, when in doubt, I'd opt for the full name topics.

Agreed. We are using `v` and `d` or `dl` for historical reasons, let's advocate full name in all other cases 👍

api/open-vsx.ts
```diff
@@ -21,28 +22,30 @@ async function handler ({ topic, pkg, namespace }: PathArgs) {
   const data = await got(endpoint).json<any>()
 
   switch (topic) {
-    case 'v':
     case 'version':
```
vladimyr, line 25:
```suggestion
	case 'v':
    case 'version':
```

vladimyr (changes requested):
> Agreed. We are using `v` and `d` or `dl` for historical reasons, let's advocate full name in all other cases :+1:

@idleberg I might be interpreting this incorrectly but I believe @amio just established a policy that says the following:
- **version** route should support both short (`v`) and long forms
- **license** route should support both short (`l`) and long forms _(my own addition to follow the suit based upon historical precedence)_
- **download count** routes should use short forms exclusively (`d` for downloads, `dt` for total downloads, `dm` for monthly downloads, and so on)
- every **other** route should use a long (kebab-case) form exclusively

According to that, I propose the following changes...

vladimyr (changes requested):
Sorry, I missed the route pattern in the previous review.
Also, the download count route should become `/d` instead of `/downloads` ie supporting only short form.

> * **download count** routes should use short forms exclusively (`d` for downloads, `dt` for total downloads, `dm` for monthly downloads, and so on)

api/open-vsx.ts
```diff
@@ -6,13 +6,14 @@ import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
...
-    '/open-vsx/d/idleberg/electron-builder': 'downloads'
+    '/open-vsx/downloads/idleberg/electron-builder': 'downloads',
+    '/open-vsx/license/idleberg/electron-builder': 'license',
+    '/open-vsx/rating/idleberg/electron-builder': 'rating',
+    '/open-vsx/reviews/idleberg/electron-builder': 'reviews',
+    '/open-vsx/version/idleberg/electron-builder': 'version'
   },
   handlers: {
-    '/open-vsx/:topic/:namespace/:pkg': handler
+    '/open-vsx/:topic<downloads|license|rating|reviews|version>/:namespace/:pkg': handler
```
vladimyr, line 16:
```suggestion
    '/open-vsx/:topic<d|l|license|rating|reviews|v|version>/:namespace/:pkg': handler
```

vladimyr:
@idleberg I feel I'm really messing up with your time :cry: 

So this time I offer you something better:

```patch
diff --git a/api/open-vsx.ts b/api/open-vsx.ts
index bf2a1b7..457dfb4 100644
--- a/api/open-vsx.ts
+++ b/api/open-vsx.ts
@@ -6,7 +6,7 @@ import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
 export default createBadgenHandler({
   title: 'Open VSX',
   examples: {
-    '/open-vsx/downloads/idleberg/electron-builder': 'downloads',
+    '/open-vsx/d/idleberg/electron-builder': 'downloads',
     '/open-vsx/license/idleberg/electron-builder': 'license',
     '/open-vsx/rating/idleberg/electron-builder': 'rating',
     '/open-vsx/reviews/idleberg/electron-builder': 'reviews',
@@ -30,7 +30,6 @@ async function handler ({ topic, pkg, namespace }: PathArgs) {
         color: versionColor(data.version)
       }
     case 'd':
-    case 'downloads':
       return {
         subject: 'downloads',
         status: millify(data.downloadCount),
@@ -55,11 +54,5 @@ async function handler ({ topic, pkg, namespace }: PathArgs) {
         status: millify(data.reviewCount),
         color: 'green'
       }
-    default:
-      return {
-        subject: 'Open VSX',
-        status: 'unknown',
-        color: 'grey'
-      }
   }
 }

```
Patch download: http://sprunge.us/btxqXj

The default case is redundant if you explicitly match all possible routes using a route pattern. :point_up:

vladimyr (approved):

amio:
Thanks! @idleberg @vladimyr

PR #513, opened 2021-03-03
Add endpoint for PyPi Python versions

NickKaramoff:
This badge displays Python versions from the Classifiers of a PyPi package.

grafik

The code is almost completely [taken from Shields.io](https://github.com/badges/shields/blob/cf7c9c11471bb227069c013657491d842f6c7940/services/pypi/pypi-python-versions.service.js) — I don't know whether that's okay or not.

Anyway, I've tested it, and it seems to work :+1:

### DIFF
--- a/api/pypi.ts
+++ b/api/pypi.ts
@@ -1,4 +1,5 @@
 import got from '../libs/got'
+import { compare, coerce } from 'semver'
 import { version, versionColor } from '../libs/utils'
 import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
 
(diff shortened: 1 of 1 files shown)


vladimyr:
Not bad but you can make it even quicker and cleaner using, in fact, less code:
```patch
diff --git a/api/pypi.ts b/api/pypi.ts
index 2d47cbb..3624904 100644
--- a/api/pypi.ts
+++ b/api/pypi.ts
@@ -1,4 +1,5 @@
 import got from '../libs/got'
+import { coerce, compare, SemVer } from 'semver'
 import { version, versionColor } from '../libs/utils'
 import { createBadgenHandler, PathArgs } from '../libs/create-badgen-handler'
 
@@ -8,9 +9,10 @@ export default createBadgenHandler({
     '/pypi/v/pip': 'version',
     '/pypi/v/docutils': 'version',
     '/pypi/license/pip': 'license',
+    '/pypi/python/black': 'python version',
   },
   handlers: {
-    '/pypi/:topic<v|license>/:project': handler
+    '/pypi/:topic<v|license|python>/:project': handler
   }
 })
 
@@ -31,5 +33,43 @@ async function handler ({ topic, project }: PathArgs) {
         status: info.license || 'unknown',
         color: 'blue'
       }
+    case 'python': {
+      const { standard, exclusive } = readVersions(info.classifiers)
+      const versions = (standard.length ? standard : exclusive).join(' | ')
+      return {
+        subject: 'python',
+        status: versions || 'unknown',
+        color: versions ? 'blue' : 'grey'
+      }
+    }
+  }
+}
+
+function readVersions(classifiers: string[]) {
+  const reVersionClassifier = /^Programming Language :: Python :: ([\d.]+)( :: Only)?$/i
+  const versions = classifiers.reduce((acc, classifier) => {
+    const match = classifier.match(reVersionClassifier)
+    if (!match) return acc
+    const [, source, isExclusive] = match
+    const version = coerce(source)
+    if (!version) return acc
+    const versionDict = isExclusive ? acc.exclusive : acc.standard
+    versionDict.delete(version.major.toString())
+    versionDict.set(source, version)
+    return acc
+  }, {
+    standard: new Map<string, SemVer>(),
+    exclusive: new Map<string, SemVer>()
+  })
+  const compareFn = (source1: string, source2: string, versionDict: Map<string, SemVer>) => {
+    return compare(versionDict.get(source1)!, versionDict.get(source2)!)
+  }
+  return {
+    standard: sortKeys(versions.standard, compareFn),
+    exclusive: sortKeys(versions.exclusive, compareFn)
   }
 }
+
+function sortKeys<K, V>(map: Map<K, V>, compareFn: (a: K, b: K, map: Map<K, V>) => number) {
+  return Array.from(map.keys()).sort((a, b) => compareFn(a, b, map))
+}

```

You can checkout commit on my fork: https://github.com/vladimyr/badgen.net/commit/60938f1

PR #520, opened 2021-04-04
docs: remove repeated word in https documentation

ajtorres9:
Removes unnecessary `then` from the second step of the "Use Badgen with HTTPS Endpoint" example


vladimyr (approved):

amio:
@ajtorres9 Thanks!

PR #523, opened 2021-04-23
Update GitHub examples

lens0021:
to recover https://github.com/badgen/badgen.net/commit/d190f787b8843155f4e04befcc23aac786c3fcb4,
which is reverted at https://github.com/badgen/badgen.net/commit/e2814bf4f470dd8f6c86bffdf75613fbe00a7bf7


amio:
It would be nice to auto detect default branch. 

For now it is just for "master" branch, thanks @lens0021 !

PR #526, opened 2021-05-12
badge(github): Use default branch for checks and status

lens0021:
This makes combined checks and combined statuses use the default branch when the branch is omitted.
It is inspired by the below comment.

> It would be nice to auto detect default branch.
> 
> For now it is just for "master" branch, thanks @lens0021 !

_Originally posted by @amio in https://github.com/badgen/badgen.net/issues/523#issuecomment-826226596_

I did not test this and it will be a breaking change, however.


amio:
@lens0021 Sorry for the late response, that's a great PR!

PR #533, opened 2021-06-29
Set desired npm version to 7

bodazhao:
As defined by the `lockfileVersion` in `package-lock.json`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,7 +56,8 @@
         "typescript": "^4.2.4"
       },
       "engines": {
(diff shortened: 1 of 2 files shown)


wopian:
Node 14 ships with npm 6.x: https://nodejs.org/en/download/releases/

bodazhao:
@wopian but the current `package-lock.json` is in `lockfileVersion` 2, meaning it was built with NPM 7

amio:
Thanks! @bodazhao

PR #536, opened 2021-08-07
Add total downloads support to NuGet

shibayan:
Add support for total downloads to the existing NuGet badge. NuGet package IDs are treated as all lowercase in the API, so I've added lowercase conversion to the existing badge.

```
/nuget/dt/:project
```

### Preview

image


amio:
Thanks! @shibayan

PR #556, opened 2022-06-09
fix(github): Cover all conclusion enums in the GitHub check-runs API

blooper05:
Hi! Thank you for a cool service 😆

I found that when I try to make some of my badges for GitHub checks, I get 500 errors.
e.g. <https://badgen.net/github/checks/blooper05/action-rails_best_practices>

I have investigated and found that the checks in those errored repositories contain the `neutral` on `conclusion` property.
I also read the documents accordingly and found that there are other `conclusion`s that are not covered by the current implementation.
ref. <https://docs.github.com/en/rest/checks/runs#list-check-runs-for-a-git-reference>

So, I have fixed these issues 👍 
Please check my changes and merge them if you like them. Thanks!

### DIFF
--- a/api/github.ts
+++ b/api/github.ts
@@ -89,14 +89,16 @@ const statesColor = {
   unknown: 'grey'
 }
 
+// https://docs.github.com/en/rest/checks/runs#list-check-runs-for-a-git-reference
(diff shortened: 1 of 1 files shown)


amio:
Thanks for the fix! @blooper05

PR #561, opened 2022-07-22
fix(docker): support OCI image spec

### LINKED ISSUE #560: Docker Metadata should support OCI Schema because Label Schema is superseded by it
The `Label Schema`[^1] is deprecated and replaced by `OCI Image Spec`[^2]. So badgen should support the `org.opencontainers.image.*` annotations.

[^1]: http://label-schema.org/rc1/
[^2]: https://github.com/opencontainers/image-spec/blob/main/annotations.md#back-compatibility-with-label-schema

viceice:
- closes #560


amio:
Thanks @viceice !

PR #562, opened 2022-07-26
fix(api/npm): scan `exports` for types

DerYeger:
Types can also be specified in the `exports` property of `package.json` files.
Since this property supports [multiple levels of nesting](https://nodejs.org/api/packages.html#nested-conditions), my proposal recursively scans them all.

### DIFF
--- a/api/npm.ts
+++ b/api/npm.ts
@@ -183,15 +183,17 @@ async function dependents (name: string) {
 async function typesDefinition(pkg: string, tag = 'latest') {
   let meta = await pkgJson(pkg, tag)
 
-  const hasExportTypes = (meta: any) => {
+  const hasTypes = (meta: any) => {
     if (typeof meta.types === 'string' || typeof meta.typings === 'string') {
       return true
     }
-    const hasNestedTypes = (exports: any) =>
-      typeof exports === 'object' &&
-      (typeof exports.types === 'string' ||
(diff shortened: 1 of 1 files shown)


api/npm.ts
```diff
@@ -183,17 +183,35 @@ async function dependents (name: string) {
...
-  if (typeof meta.types === 'string' || typeof meta.typings === "string") {
+  const hasExportTypes = (meta: any) => {
+    if (typeof meta.types === 'string' || typeof meta.typings === 'string') {
+      return true
+    }
+    const hasNestedTypes = (exports: any) =>
+      typeof exports === 'object' &&
+      (typeof exports.types === 'string' ||
+      Object.values(exports)
+        .some(hasNestedTypes))
```
amio, line 194: These formating seems a bit confusing, how about this:
```js
const hasNestedTypes = (exports: any) => {
  if (typeof exports !== 'object') return false
  if (typeof exports.types === 'string') return true
  if (Object.values(exports).some(hasNestedTypes)) return true
  return false
}
```

DerYeger: That's much better indeed!

DerYeger:
Is there anything blocking this PR, which I should change?

amio:
Sorry for missing the updates, and thanks @DerYeger !

PR #563, opened 2022-08-05
Fix issue when loading platform/sdk info from pub.dev

d-markey:
The response from pub.dev's `api/packages/${pck}/metrics` API has changed.
`derivedTags` is now available from `$.scorecard.panaReport.derivedTags`


amio:
@d-markey Thanks for the fix!

PR #579, opened 2022-12-31
feat: upgrade next to v13 (WIP)

amio:
New year, new beginning

- [x] upgrade nextjs to v13
- [x] new home page
- [x] rename static badge generator `/badge` -> `/static`
- [x] overhaul error tracking with sentry
- [ ] migrate badge generators
  - [x] github
  - [x] winget
  - [x] xo

PR #588, opened 2023-02-06
apm: discontinue service

idleberg:
Following this [discussion](https://github.com/badgen/badgen.net/discussions/584), I'm removing support for `apm`, the package manager for the Atom editor.

Question: Should the service remain listed in `badge-list.ts`?


amio:
We should remove it from badge-list.

PS, I just merged this into #579

idleberg:
@amio Do I have to take any action to trigger the failed deployment pipeline?

PR #595, opened 2023-06-06
feat: add pypi downloads

vladkens:
Hi. I added routes to show packages downloads in pypi.
```bash
/pypi/dd/pip
/pypi/dw/pip
/pypi/dm/pip
```

image


amio:
Hi @vladkens thanks for the PR! I just migrated the changes into https://github.com/badgen/badgen.net/pull/599

amio:
It's deployed, you may checkout https://badgen.net/pypi

PR #659, opened 2024-02-20
Fix homepage typo

DominikPeters:
"generater" should be "generator", see
https://www.merriam-webster.com/dictionary/generator
https://dictionary.cambridge.org/dictionary/english/generator


amio:
Thanks @DominikPeters !

PR #674, opened 2024-07-17
fix: Change simpleicons twitter --> x to resolve icon render

Spiteless:
## `Fix`
Image src coming from `simpleicons`  for `/twitter`  wasn't reendering a twitter icon in the `serve-doc` and `serve-doc-next` footer.

<details>
<summary>Screenshots</summary>

### Before

image

### After

image

</details>


amio:
Yep, we should update it

PR #675, opened 2024-07-17
fix: Update footer in serve doc

Spiteless:
## `Fix`
`serve-doc` and `serve-doc-next` footer differed from the main `footer.tsx` file in styling and anchor links.

Updated the CSS to match as best I could. They're not the exact same but they're close.

<details>
<summary>Screenshots</summary>

** Disregard the weird color of the heart emoji in `with ♥ by`, something isn't rendering right on my laptop, it renders fine on a phone.

### Before

`https://badgen.net/memo`
image

### After

`http://localhost:3000/memo`
image

### Favicons

`Before: left` `After: right`
image

</details>

### DIFF
--- a/libs/serve-doc-next.ts
+++ b/libs/serve-doc-next.ts
@@ -120,11 +120,10 @@ const helpFooter = `
   </div>
   <style>
     footer {
-      margin-top: 5rem;
       background-color: #222;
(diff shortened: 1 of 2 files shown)


amio:
Nice update, thanks @Spiteless!

PR #712, opened 2025-08-09
Use BadgenError from Next handler

Falke-Design:
I'm not sure when the Next-handler and when the legacy handler are used, but when I run it locally, it uses the Next-handler. 

When for example no Github token is set, it doesn't use the correct `BadgenError` from `create-badgen-handler-next` which causes the handler to fail because `error.message` is undefined and that is not allowed for `setHeader`.

https://github.com/badgen/badgen.net/blob/861a20a29de85382b26b17f5fc911620a4853457/libs/create-badgen-handler-next.ts#L104

https://github.com/badgen/badgen.net/blob/861a20a29de85382b26b17f5fc911620a4853457/libs/github.ts#L33

### DIFF
--- a/libs/docker.ts
+++ b/libs/docker.ts
@@ -1,5 +1,5 @@
 import got from './got'
-import { BadgenError } from './create-badgen-handler'
+import { BadgenError } from './create-badgen-handler-next'
 
 // request image specific DockerHub pull token
 export function getDockerAuthToken<T = any>(scope: string, name: string) {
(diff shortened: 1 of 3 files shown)


amio:
@Falke-Design Thanks for the finding!

The "next-handler" is the handler we should use, the plan is to migrate all legacy endpoints to new handler one-by-one, so some legacy code are still live in `/api-legacy` but actually they should not serve any requests. All requests to legacy handler will be redirected to https://v2022.badgen.net respectively.

PR #713, opened 2025-08-09
Add env vars to Readme

Falke-Design:
I added missing env vars to the Readme


amio:
Triple thanks! @Falke-Design

Issue #611, opened 2023-06-27
Some flat badges are not flat


bokub:
Hi,

I have noticed that some "flat" badges are grey & round, such as "codeclimate" or "codecov" badges:

[https://flat.badgen.net/codeclimate](https://flat.badgen.net/codeclimate) 
[https://flat.badgen.net/codecov](https://flat.badgen.net/codecov) 

![Badge](https://flat.badgen.net/codeclimate/maintainability/codeclimate/codeclimate) ![Badge](https://flat.badgen.net/codecov/c/github/babel/babel)

Correct flat badge for comparison:

![Badge](https://flat.badgen.net/badge/color/green/green)

amio:
Oh true, the migration recently just proxy some badge to legacy instance, but forgot to handle the badge style switch.

bokub:
Looks like it's fixed now :+1: 

Thanks

bokub:
Hi @amio 

It looks like the j[sdelivr badges](https://flat.badgen.net/jsdelivr) are affected by this issue too, so I'm reopening it

![image](https://github.com/user-attachments/assets/50bcd0f0-4d42-4aaa-9a1a-348f2ef17ba4)

PR #715, opened 2025-08-09
Fix jsDelivr flat icon

Falke-Design:
Fixes: #611


amio:
Thanks! @Falke-Design

PR #716, opened 2025-08-09
Fix liberapay flat badges

Falke-Design:
https://flat.badgen.net/liberapay

grafik

Followup: #611


amio:
@Falke-Design  Thanks!

PR #781, opened 2026-06-18
refactor: millify Chrome Web Store user count

crimx:
Format Chrome Web Store user counts with `millify`, so values render consistently with other count badges.

I also noticed that `.nvmrc` is still set to Node.js 18, while `package.json` requires Node.js `>=24`. Should `.nvmrc` be updated to match the engine requirement?


amio:
Yes the `.nvmrc` needs to be updated.

The vercel authorization page keep showing error now, maybe I need to try another day.

